PR #10, opened 2020-01-15
Spline #554 better sanity check messages

cerveada:


wajda:
Fixes absaoss/spline#554

PR #11, opened 2020-01-28
Scala cross-build support

wajda:


wajda:
Fixes AbsaOSS/Spline#212

PR #18, opened 2020-04-01
Spline #615 Http timeouts and request compression

cerveada:
https://github.com/absaoss/spline/issues/615

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
@@ -16,20 +16,30 @@
 package za.co.absa.spline.harvester.dispatcher
 
 
+import java.io.ByteArrayOutputStream
+import java.util.zip.GZIPOutputStream
+
 import org.apache.commons.configuration.Configuration
 import org.apache.spark.internal.Logging
 import scalaj.http.{BaseHttp, Http}
 import za.co.absa.commons.config.ConfigurationImplicits._
 import za.co.absa.spline.harvester.dispatcher.HttpLineageDispatcher.RESTResource
+import za.co.absa.spline.harvester.dispatcher.HttpLineageDispatcher._
 import za.co.absa.spline.harvester.exception.SplineNotInitializedException
 import za.co.absa.spline.harvester.json.HarvesterJsonSerDe.impl._
 import za.co.absa.spline.producer.model.{ExecutionEvent, ExecutionPlan}
+import za.co.absa.commons.lang.ARM.{managed, using}
 
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
 object HttpLineageDispatcher {
   val producerUrlProperty = "spline.producer.url"
+  val connectionTimeoutMsKey = "spline.timeout.connection"
+  val readTimeoutMsKey = "spline.timeout.read"
+
+  val defaultConnectionTimeout = 1000
+  val defaultReadTimeout = 20000
 
   object RESTResource {
     val ExecutionPlans = "execution-plans"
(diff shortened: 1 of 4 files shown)


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -16,20 +16,30 @@
...
 import za.co.absa.spline.harvester.json.HarvesterJsonSerDe.impl._
 import za.co.absa.spline.producer.model.{ExecutionEvent, ExecutionPlan}
+import za.co.absa.commons.lang.ARM.{managed, using}
 
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
 object HttpLineageDispatcher {
   val producerUrlProperty = "spline.producer.url"
+  val connectionTimeoutMsKey = "spline.timeout.connection"
```
wajda, line 38: Constant names should be in upper camel case

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -61,20 +98,41 @@ class HttpLineageDispatcher(splineServerRESTEndpointBaseURL: String, http: BaseH
 
   private def sendJson(json: String, url: String) = {
     log.debug(s"sendJson $url : $json")
-    try http(url)
-      .postData(json)
-      .compress(true)
+
+    try {
+      val http = if (useRequestCompression) {
```
wajda, line 103: nit: try to keep `if` and `else` keywords vertically aligned. It's much easier to read.
In the current code block `else` is aligned with `val`, rather then `if`.   
E.g.
```scala
val http = 
  if (...) {
    ...
  } else { // aligned with 'if'
    ...
  } 
```
vs
```scala
val http = if (...) {
    ...
  } else { // aligned with 'http'
    ...
  } 
```

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -61,20 +98,41 @@ class HttpLineageDispatcher(splineServerRESTEndpointBaseURL: String, http: BaseH
...
+        baseHttp(url)
+          .postData(compressData(json))
+          .header("content-encoding", "gzip")
+      } else {
+        baseHttp(url)
```
wajda, line 108: Repeating `baseHttp(url)` could be extracted to a variable.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -87,4 +145,33 @@ class HttpLineageDispatcher(splineServerRESTEndpointBaseURL: String, http: BaseH
...
+      .throwError
+      .body)
+
+    tryStatus
+      .toOption
+      .map(extractCompressionFlag(_))
+      .getOrElse(false)
+  }
+
+  private def extractCompressionFlag(json: String): Boolean = {
```
wajda, line 162: Can we use JSON commons for this?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -61,20 +98,41 @@ class HttpLineageDispatcher(splineServerRESTEndpointBaseURL: String, http: BaseH
...
       .body
-    catch {
+
+    } catch {
       case NonFatal(e) => throw new RuntimeException(s"Cannot send lineage data to $url", e)
     }
   }
 
+  private def compressData(json: String): Array[Byte] = {
+
```
wajda, line 126: nit: unnecessary empty lines

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -87,4 +145,33 @@ class HttpLineageDispatcher(splineServerRESTEndpointBaseURL: String, http: BaseH
       case _ => Unit
     }
   }
+
+  private def isServerCompressionReady(): Boolean = {
```
wajda, line 149: Why not combining this with the `ensureProducerReady` method? It seems like a duplicated HTTP call.

cerveada: To be backwards compatible we need to support HEAD because GET may fail. We could use HEAD only as a fallback though.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
cerveada: This is such a simple call that It seems to me it would only increase the cognitive complexity without any real benefits.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
cerveada: I would like this to work without need for specific format of the incoming JSON, that would make it forward compatible when we add more fields.

wajda: `JsonSerDe` can do it as well, and it's a one-liner.
See a usage example in a test here - https://github.com/AbsaOSS/commons/blob/master/src/test/scala/za/co/absa/commons/json/JsonSerDeSpec.scala#L63

wajda:
```scala
val isCompressionSupported = body.fromJson[Map[String, Boolean]].getOrElse("requestDecompression", false)
```

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
wajda: But we still can do it in one place, right? We try GET first and recover with HEAD. I don't see any reason why should we postpone the call till the time we are about sending captured lineage data out.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
wajda: Ok. Just fix indentation inside `try-block`, and in the class constructor signature please.

PR #22, opened 2020-04-07
Issue #605

radford1:
fixes AbsaOSS/spline#605


integration-tests/pom.xml
```diff
@@ -102,12 +102,24 @@
...
         <dependency>
             <groupId>org.cassandraunit</groupId>
             <artifactId>cassandra-unit</artifactId>
             <version>2.1.9.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>de.flapdoodle.embed</groupId>
+            <artifactId>de.flapdoodle.embed.mongo</artifactId>
+            <version>1.47.1</version>
```
wajda, line 120: There is no sense in adding `<version>` here if you have it in the `<dependencyManagement>` section below.

Otherwise LGTM.

radford1: I'm not sure I am following. I think this is the only spot that the flapdoodle.embed.mongo is referenced? Should I move the version reference to the project level pom.xml?

wajda: Oh, it was for the `mongo-spark-connector_${scala.binary.version}`, line #108. Sorry

radford1: Ah got ya. I'll update that and the cassandra connector which has the same issue

core/src/main/scala/za/co/absa/spline/harvester/builder/read/ReadCommandExtractor.scala
```diff
@@ -82,11 +82,26 @@ class ReadCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
...
           ReadCommand(SourceIdentifier.forCassandra(keyspace.asInstanceOf[String], table.asInstanceOf[String]), operation)
 
+        case `_: MongoDBSourceRelation`(mongr) =>
+          val mongoRDD = extractFieldValue[AnyRef](mongr, "mongoRDD")
+          val readConfig = extractFieldValue[AnyRef](mongoRDD, "readConfig")
+          val database = extractFieldValue[AnyRef](readConfig, "databaseName")
+          val collection = extractFieldValue[AnyRef](readConfig, "collectionName")
+          val connectionUrlx = extractFieldValue[AnyRef](readConfig, "connectionString")
+
+          val connectionUrl_Scala_2_11 = "x"
```
wajda, line 97: This slipped out of my sight. Those `x` and `value`, aren't you are dealing with `Option` monad here?
Can we use more strict types instead of `AnyRef` that would be compatible between Spark versions?

wajda: Please take a look at my commit here. Do you agree with that? 4bf769f1e78a2edca94dbd86cc7129c0267ef29b

wajda:
Closing this PR in favor of #24 which is a re-based version of this one + minor fixes

core/src/main/scala/za/co/absa/spline/harvester/builder/read/ReadCommandExtractor.scala
radford1: Yes got it. I'll try and implement future ones like this instead of the way I was doing it.

PR #25, opened 2020-04-10
added elasticsearch support issue-#606

radford1:


radford1:
@wajda is it possible to get the rat file to inspect from the build server? /mnt/agent/work/ed446036424b4f6e/integration-tests/target/rat.txt

wajda:
Yes, sure. It's stored as a log artifact in a TeamCIty build. See https://teamcity.jetbrains.com/buildConfiguration/OpenSourceProjects_AbsaOSS_SplineAgentSpark_PullRequestsSpark22scala211/2863275?buildTab=artifacts

pom.xml
```diff
@@ -95,6 +95,9 @@
         <scala_2.11.version>2.11.12</scala_2.11.version>
         <scala_2.12.version>2.12.10</scala_2.12.version>
 
+        <!-- Unsupported connectors -->
+        <exlude.unsupported>None</exlude.unsupported>
```
wajda, line 99: We have a conditional ignore tag for this purpose. It's much easier to use. https://github.com/AbsaOSS/spline/issues/606#issuecomment-613321957

core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -54,13 +54,18 @@ class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
           case Some(sourceType) if sourceType == "mongo" || sourceType.isInstanceOf[com.mongodb.spark.sql.DefaultSource] =>
             asMongoDBWriteCommand(cmd)
 
+          case Some(sourceType) if sourceType == "elasticsearch" || sourceType.isInstanceOf[org.elasticsearch.spark.sql.DefaultSource15] =>
```
wajda, line 57: I just noticed the usage if `.isInstanceOf` here as well as for Mongo and Cassandra cases. Although it works in tests, but will break in production due to missing classes in runtime. `.isInstanceOf[A]` requires class `A` to be loaded into JVM.
Instead we need to utilize `SafeTypeMatchingExtractor` - the same approach that is used for Excel type checking (see line #60). Similar to what you do in the `ReadCommandExtractor` class.

radford1: Ok I'll take a look and try to correct all 3

PR #26, opened 2020-04-23
Issue #642

radford1:
Fixes AbsaOSS/spline#642

Prior to spark 2.4 the databricks avro connector was required. I coded to support both databricks connector prior to spark 2.4 and spark avro in spark 2.4


radford1:
The duplication report is around multiple test cases testing different file formats. Let me know if this is not ok

radford1:
@wajda does the test case need to be broken out into separate files in order to pass duplication?

wajda:
Hi @radford1 , sorry for keeping you waiting. Being buried under work recently having Spline 05 release done. Will take a look at your PR later today.
In general I'd say no, you don't have to. I don't mind duplication in tests. Clarity is much more important. On the other hand some general purpose fixtures could be created to eliminate the majority of code duplicates. For now it's up to you. We could make some refactoring in tests in the future releases.

radford1:
Hey @wajda I would be comfortable leaving the test case as is so that it is clear what it is testing.

core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -74,9 +79,18 @@ class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
 
       case cmd: InsertIntoHadoopFsRelationCommand =>
         val path = cmd.outputPath.toString
-        val format = cmd.fileFormat.toString
         val qPath = pathQualifier.qualify(path)
-        WriteCommand(cmd.nodeName, SourceIdentifier(Some(format), qPath), cmd.mode, cmd.query, cmd.options)
+        val fileFormat = cmd.fileFormat
+        fileFormat match {
+          case `_: SparkAvroSourceExtractor`(avro) =>
```
wajda, line 85: Naming convention:
This notation ``` `_: Something` ``` supposes to simulate Scala type matching pattern, that you would normally use if the type `Something` were guaranteed to be available in the runtime. In this case we don't use `Extractor` or any other suffix to the type name, to make it as close to the native Scala syntax as possible. This is the only case we use that "weird" notation. So either remove the suffix or the back ticks.

wajda: Otherwise LGTM. Thanks.

PR #67, opened 2020-05-12
Spark agent #39 ignore data from LocalRelation

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/converter/OperationParamsConverter.scala
```diff
@@ -60,5 +60,5 @@ class OperationParamsConverter(
 }
 
 object OperationParamsConverter {
-  private val knownPropNames = Set("nodeName", "output", "children")
+  private val knownPropNames = Set("nodeName", "output", "children", "data")
```
wajda, line 63: Is `data` property guaranteed to have the same semantics for every operation?
`knownPropNames` supposes to contain the properties that are common for all operations with clearly defined semantics that we process directly, and thus need to be ignored by the reflective traversing.

cerveada: Not really, but I searched the testing database (from Sasa) and it doesn't seem to be used anywhere else there. The advantage of this is that if there is other operation than LocalRelation that contains user data it will be ignored as well.

wajda: Ok, makes sense. But just for the sake of clarity I would create another constant, say `ignoredPropNames` and put `"data"` there.

Issue #74, opened 2020-05-26
Lineage not being reported from production Hadoop+Hive+Spark


DaimonPl:
I verified that lineage is reported from end-to-end tests (local spark context)

However when run on hadoop cluster with Hive and Spark 2.4.5, no lineage, nor any warning or error is reported

configuration:

```
spark.sql.queryExecutionListeners="za.co.absa.spline.harvester.listener.SplineQueryExecutionListener"
spark.spline.producer.url="http://my-domian-name:9090/producer"
spark.spline.timeout.read=30000
spark.spline.timeout.connection=30000
spark.spline.iwd_strategy.default.on_missing_metrics="CAPTURE_LINEAGE"
```

build.sbt:

```
"za.co.absa.spline.agent.spark" %% "agent-core" % "0.5.2" excludeAll(
  ExclusionRule(organization = "org.apache.spark")
)
```

logs from job:

```
2020-05-26 16:22:35.793 | Spline successfully initialized. Spark Lineage tracking is ENABLED.
2020-05-26 16:22:35.728 | spline.timeout.connection is set to:'30000 milliseconds' ms
2020-05-26 16:22:35.728 | spline.timeout.read is set to:'30000 milliseconds' ms
2020-05-26 16:22:35.727 | spline.producer.url is set to:'http://my-domian-name:9090/producer'
2020-05-26 16:22:35.709 | Instantiating LineageDispatcher for class name: za.co.absa.spline.harvester.dispatcher.HttpLineageDispatcher
2020-05-26 16:22:35.706 | Instantiating IgnoredWriteDetectionStrategy for class name: za.co.absa.spline.harvester.write_detection.DefaultIgnoredWriteDetectionStrategy
2020-05-26 16:22:35.703 | Spline mode: BEST_EFFORT
2020-05-26 16:22:35.703 | Spline version: null
2020-05-26 16:22:35.685 | Initializing Spline agent...
```

Same code run from local spark context sends lineage successfuly and it's visible in spline UI

Any idea how to debug problem? I see no logs which could be helpful, i set logging configuration for DEBUG

wajda:
It could be caused by a different event type sent by Spark when in Hive environment.
Can you reproduce it locally having Hive enabled?
[...]

PR #80, opened 2020-05-28
spline #74 Add more http debug logs

DaimonPl:


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -147,9 +147,16 @@ class HttpLineageDispatcher(restClient: RestClient)
         .post(json, requestCompressionSupported)
         .throwError
         .body
-
     } catch {
-      case NonFatal(e) => throw new RuntimeException(s"Cannot send lineage data to ${endpoint.request.url}", e)
+      case e: HttpStatusException => {
+        log.error(s"Could not send lineage data to ${endpoint.request.url}: ${e.code} ${e.statusLine}", e)
+        log.debug(s"Error response body: ${e.body}")
+        throw new RuntimeException(s"Cannot send lineage data to ${endpoint.request.url}", e)
```
wajda, line 154: Logging and throwing at the same time is considered an anti-pattern.
Throwing means giving up and delegating the error handling responsibility to the caller. In this case you don't know if it's actually an error at all, because you don't know how the calling is going to handle it. 

If the intention is to log more details on exception then the better approach would be either adding that info into the exception message, or (if you want to be picky and type safe) creating another exception class with corresponding fields.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -147,9 +147,16 @@ class HttpLineageDispatcher(restClient: RestClient)
...
+      }
+      case NonFatal(e) => {
+        log.error(s"Could not send lineage data to ${endpoint.request.url}", e)
+        throw new RuntimeException(s"Cannot send lineage data to ${endpoint.request.url}", e)
```
wajda, line 158: Same as above

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
DaimonPl: that's great example when trying to avoid "antipattern" caused several days of debugging instead of 5 minutes just because this log wasn't here :)

DaimonPl: i.e. that exception was handled by spark by just logging it as warning with message completely not related to spline - it was very easy to miss it because it was not logged by spline at all

I added that log here completely intentionally because of this reason - i still think that it will make life easier for future users if they have to debug some problems

wajda: Sorry I cannot agree with this statement. It's wasn't avoiding anti-patterns that caused you several days of debugging, but missing logs. I'm not again adding logs, don't get me wrong. My point is to do it in a right way.

DaimonPl: I've added logs to spline logger which is expected to log problems about http communication `HttpLineageDispatcher`

Adding message details to exception itself would not help at all because it was caught by spark internals and logged as spark internals logs thus making it more difficult to filter all logs from spline - that's exactly what caused those problems, spline did not log communication error details at all, extending exception message won't help at all, catching HTTP exceptions in different place also not the best idea, not throwing exception even worse - thus i still think that's the best way to have this situation improved - we have exact HTTP problem logs in http related component and still throwing exception as unexpected situation

I have no plans to make changes to this PR, so if you don't agree we can just close it, but it will be still difficult to debug http communication from agent

As I can see i'm not the only one with problems https://github.com/AbsaOSS/spline/issues/683#issuecomment-634762779

wajda: thank you @DaimonPl for you contribution. I'll give it another look.

wajda:
closed in favor of #82
@DaimonPl , could you test it on your environment please?

Issue #83, opened 2020-05-28
Add Table Catalog params for non hive tables


radford1:
When spark is using a hive metastore and the table type is using parquet instead of hive then the table catalog params are not populated.

DaimonPl:
Looks like i had same issue but thought that cause was different: https://github.com/AbsaOSS/spline/issues/685

radford1:
Ah got ya. I'm working on a fix that I'm getting ready to push that will add these to the params the same way hive tables work. Hopefully this will also help to solve your issue as well.

DaimonPl:
@wajda any chance to have 0.5.3 released soon? I'd like to deploy spark agents for 3 teams and check how everything is working - so far 0.5.2 would fail quite fast

wajda:
The release procedure started - https://teamcity.jetbrains.com/buildConfiguration/OpenSourceProjects_AbsaOSS_SplineAgentSpark_Release?mode=builds

wajda:
done

PR #84, opened 2020-05-28
Issue #83

radford1:


core/src/main/scala/za/co/absa/spline/harvester/builder/read/ReadCommandExtractor.scala
```diff
@@ -49,16 +49,22 @@ class ReadCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
...
-          fileFormat match {
-            case SparkAvroSourceRelation(avro) =>
-              ReadCommand(SourceIdentifier(Some("Avro"), uris: _*), operation, hr.options)
-            case DatabricksAvroSourceRelation(avro) =>
-              ReadCommand(SourceIdentifier(Some("Avro"), uris: _*), operation, hr.options)
-            case _ =>
-              val format = fileFormat.toString
-              ReadCommand(SourceIdentifier(Some(format), uris: _*), operation, hr.options)
+          val catalogTable = lr.catalogTable.orNull
+          if(catalogTable != null){
```
wajda, line 53: nit: missing space after `if` and before `{`

core/src/main/scala/za/co/absa/spline/harvester/builder/read/ReadCommandExtractor.scala
```diff
@@ -49,16 +49,22 @@ class ReadCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
...
+            ReadCommand(SourceIdentifier.forTable(catalogTable)(pathQualifier, session), operation, params = Map("table" -> catalogTable))
+          }
+          else{
```
wajda, line 56: nit: space before `{`

core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -78,17 +78,24 @@ class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
...
-          case DatabricksAvroSourceExtractor(avro) =>
-            WriteCommand(cmd.nodeName, SourceIdentifier(Some("Avro"), qPath), cmd.mode, cmd.query, cmd.options)
-          case _ =>
-            val format = fileFormat.toString
-            WriteCommand(cmd.nodeName, SourceIdentifier(Some(format), qPath), cmd.mode, cmd.query, cmd.options)
+        val catalogTable = cmd.catalogTable.orNull
+        if(catalogTable != null){
+          val mode = if (cmd.mode == SaveMode.Overwrite) Overwrite else Append
+          asTableWriteCommand(cmd.nodeName, catalogTable, mode, cmd.query)
+        }
```
wajda, line 85: Minor suggestion:
Since `catalogTable` is an `Option` monad, there is no need to deal with `null`s.
```scala
        cmd.catalogTable
          .map(catalogTable => {
            val mode = if (cmd.mode == SaveMode.Overwrite) Overwrite else Append
            asTableWriteCommand(cmd.nodeName, catalogTable, mode, cmd.query)
          })
          .getOrElse({
            ...
          })
```

radford1: Thanks for this suggestion. I'll try to implement this format instead of the if conditions.

Will also work on getting more encompassing test cases together. There is already an original one started called InsertIntoTest that I will expand more.

wajda:
This is what is captured in the write operation params:
```json
      "params": {
        "table": {
          "identifier": {
            "table": "path_archive_parquetserde",
            "database": "test"
          },
          "tableType": {
            "name": "MANAGED"
          },
          "storage": {
            "locationUri": {},
            "inputFormat": "org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat",
            "outputFormat": "org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat",
            "serde": "org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe",
            "compressed": false,
            "properties": {
              "serialization.format": "1"
            }
          },
          "schema": [
            {
              "name": "x",
              "dataType": {},
              "nullable": true,
              "metadata": {
                "map": {}
              }
            },
            {
              "name": "ymd",
              "dataType": {},
              "nullable": true,
              "metadata": {
                "map": {}
              }
            }
          ],
          "provider": "hive",
          "partitionColumnNames": [],
          "owner": "wajda",
          "createTime": 1590745439000,
          "lastAccessTime": 0,
          "createVersion": "2.4.2",
          "properties": {
            "transient_lastDdlTime": "1590745439"
          },
          "unsupportedFeatures": [],
          "tracksPartitionsInCatalog": true,
          "schemaPreservesCase": true,
          "ignoredProperties": {}
        }
      }
```

Do we really need to keep all that info? Schema can be huge for some jobs (thousands of columns of complex type each), and is completely redundant in this place. Wouldn't it be enough to only add `TableIdentifier` instance (and potentially a few other selected props) instead of an entire `CatalogTable`?

wajda:
Also if we want to release it quickly in the next 0.5 bugfix release it has to be based on _release/0.5_ branch.

PR #106, opened 2020-07-16
Add secondary dispatcher mechanism

schaloner-kbc:
Provide a mechanism to add secndary dispatchers that function as event listeners.  Return values from these secondary dispatchers do not interfere from the existing dispatcher mechanism and interactions.

One additional dispatcher has been provided to write into HDFS, providing backwards compatibility with Spline 0.3 functionality.

### DIFF
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -33,3 +33,5 @@ spline.iwd_strategy.default.on_missing_metrics=IGNORE_LINEAGE
 # User extra metadata provider
 spline.user_extra_meta_provider.className=za.co.absa.spline.harvester.extra.NoopUserExtraMetaDataProvider
 
+# Proprietary relation support
+spline.read_relation_handler.className=za.co.absa.spline.harvester.builder.read.NoOpReadRelationHandler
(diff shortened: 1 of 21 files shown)


cerveada:
Thank you for contribution. 

The builds are failing - probably missing licence again? Also, the year in the license is supposed to be the year when the file was created, so currently it's 2020.

wajda:
Hi Steve,
Thanks for your contribution! It certainly seems to be useful feature, but I would like to make a few notes about it.
First of all, before creating a PR we require an issue to be created first and then discussed. So that we can agree on the way how it's going to be implemented and also decide which branch/version it should go to. It's a pity you haven't done so because I believe the same thing can be achieved without a need to introduce primary and secondary dispatchers. For the simplest use case it's enough to instantiate `HttpLineageDispatcher` inside your own dispatcher and delegate to it explicitly. No changes in the agent source code are required.
If you want to do it in a more cohesive way, e.g. using a _decorator_ pattern then `AggregateLineageDispatcher` can play the role of a composite decorator, delegating to a set of other dispatchers, similar how it was done in the [`ParallelCompositeFactory`](https://github.com/AbsaOSS/spline/blob/release/0.3/persistence/api/src/main/scala/za/co/absa/spline/persistence/api/composition/ParallelCompositeFactory.scala) in the older Spline 0.3. Again, no changes in the current agent code would be required except from adding one new class and a test.

Issue #69, opened 2020-05-16
Security issue: JDBC URI contains plaintext password & username


roman-novosad:
## Proposal
Spline Agent should contain URI filtering for, at least, passwords. Captured Lineage data are send insecurely thru HTTP protocol to Producer, which can be located out of the secured network, so anyone can listen and capture the password. 

I believe this issue should be solved on either Agent and UI sides. Best solution is not sent it at all or securely mask it (f.e. with asterisks). 

## Example
Current state:
`DataSource URI: jdbc:sqlserver://sample.database.windows.net:1433;database=sample;user=sample;password=password;encrypt=true;trustServerCertificate=false;hostNameInCertificate=*.database.windows.net;loginTimeout=30:
`

Future state:
`DataSource URI: jdbc:sqlserver://sample.database.windows.net:1433;database=sample;user=;password=;encrypt=true;trustServerCertificate=false;hostNameInCertificate=*.database.windows.net;loginTimeout=30:
`
## Version 
For version 0.5.1 Agent, UI.

radford1:
This might be a little difficult to solve as a blanket fix. Most drivers have the variable as password=<password> so that would be easy to catch and replace with **** but I'm not sure if all jdbc drivers have this exact parameter name as password=<password>. Would it be ok to implement a small method that will take the url and start by replacing password=<password> with password=****. If new variations of password come up we can implement more logic in that method.

wajda:
Very good point. JDBC standard is indeed too flexible in terms ways to pass credentials to the driver.
Here's a related StackOverflow question - https://stackoverflow.com/questions/5547224/username-password-in-jdbc-connection-url

The problem here is a bit more general than just hiding JDBC passwords.
Ideally a data source URI should only contain parts that identify the location (no auth credentials, no other extra connection parameters, protocol variations etc). Because this is the primary goal for the datasource URI - to identify (uniquely and consistently) a data location in the network. Unfortunately many storage providers use URLs or connection strings to not only identify a location but also pass a bunch of other auth and configuration stuff to the driver.

[...]

PR #139, opened 2020-10-15
#69 add secret filtering feature

cerveada:


core/src/main/scala/za/co/absa/commons/CaptureGroupReplacer.scala
```diff
@@ -0,0 +1,52 @@
...
+      str
+    else {
+      val nextStr = regexes.head
+        .findFirstMatchIn(str)
+        .map(regMatch => replaceMatchedGroups(regMatch, str))
+        .getOrElse(str)
+
+      replace(nextStr, regexes.tail)
+    }
+  }
```
wajda: Decomposing a collection with pattern-matching is more idiomatic and readable than `.isEmpty`, `.head` and `.tail` combination.
Also `@tailrec` makes a good fit here.
```suggestion
  @tailrec
  final def replace(str: String, regexes: Seq[Regex]): String = regexes match {
    case Nil => str
    case regex :: nextRegexes =>
      val nextStr = regex
        .findFirstMatchIn(str)
        .map(regMatch => replaceMatchedGroups(regMatch, str))
        .getOrElse(str)
      replace(nextStr, nextRegexes)
  }
```

core/src/main/scala/za/co/absa/commons/CaptureGroupReplacer.scala
```diff
@@ -0,0 +1,52 @@
...
+    val intervals = (start +: groupsStartsAndEnds :+ end)
+      .sliding(2,1)
+      .map{ case (_, firstEnd) :: (secondStart, _) :: Nil => (firstEnd, secondStart)}
+      .toSeq
+
+    intervals
+      .map{ case (from, to) => str.substring(from, to)}
+      .mkString(replacement)
+  }
+}
```
wajda: nit: empty line

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -95,14 +100,17 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
   protected def ignoredWriteDetectionStrategy: IgnoredWriteDetectionStrategy = instantiate[IgnoredWriteDetectionStrategy](
     configuration.getRequiredString(IgnoreWriteDetectionStrategyClass))
 
+  protected def postProcessingFilters: Seq[Filter] = configuration.getRequiredString(postProcessingFilterClasses)
+    .split(";").map(_.trim).map(instantiate[Filter])
+
```
wajda: 1. A method chain is much easier to read/debug/maintain when arranged one method per line.
1. Method `split()` doesn't de-duplicate delimiters and can return empty strings. Hence an additional `filter(_.nonEmpty)` is required.
1. Usually in Java a comma is widely used as a value delimiter. If done so then the method `getRequiredStringArray()` could be used instead of explicit splitting and trimming.

```suggestion
  protected def postProcessingFilters: Seq[Filter] =
    configuration
      .getRequiredStringArray(postProcessingFilterClasses)
      .filter(_.nonEmpty)
      .map(instantiate[Filter])
```

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -53,9 +53,14 @@ object DefaultSplineConfigurer {
...
     /**
-     * Which strategy should be used to detect mode=ignore writes
+     * Class allowing to add extra info to generated lineage (Deprecated: use filter instead)
      */
     val UserExtraMetadataProviderClass = "spline.user_extra_meta_provider.className"
+
+    /**
+     * User defined filters: allowing modification and enrichment of generated lineage
+     */
+    val postProcessingFilterClasses = "spline.postprocessing_filter.classNames"
```
wajda: A constant name should start with a capital letter.

core/src/main/scala/za/co/absa/spline/harvester/extra/UserExtraAppendingFilter.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import za.co.absa.spline.harvester.HarvestingContext
+import za.co.absa.spline.harvester.postprocessing.Filter
+import za.co.absa.spline.producer.model.v1_1._
+import za.co.absa.spline.harvester.ExtraMetadataImplicits._
+
+/**
+ * provides backward compatibility for older less powerful UserExtraMetadataProvider API
+ */
+class UserExtraAppendingFilter(userExtraMetadataProvider: UserExtraMetadataProvider) extends Filter {
```
wajda: Shouldn't it be annotated with `@deprecated`?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/DataSourceUriFilter.scala
```diff
@@ -0,0 +1,41 @@
...
+  def this(conf: Configuration) = this()
+
+  override def processReadOperation(op: ReadOperation, ctx: HarvestingContext): ReadOperation =
+    op.copy(inputSources = op.inputSources.map(filter))
+
+  override def processWriteOperation(op: WriteOperation, ctx: HarvestingContext): WriteOperation =
+    op.copy(outputSource = filter(op.outputSource))
+
+  protected val replacement = "*****"
+  protected val regex = """password=([^;\s]+)""".r //NOSONAR
```
wajda: Unless there is a specific reason, usually constants are placed in a companion object for better visibility.

Also why these members are `protected`? If it's for potential extending then I'd suggest to pass them as constructor parameters instead (OCP).

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/Filter.scala
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.postprocessing
+
+import za.co.absa.spline.harvester.HarvestingContext
+import za.co.absa.spline.producer.model.v1_1._
+
+trait Filter {
```
wajda: The word "Filter" is too generic. Think about using more specific name for this.
Also why is it a filter? A word "processor" or "handler" seems to fit better.
What about "PostProcessor"? And the class that you named as "PostProcessor" could be renamed to "CompositPostProcessor" for example. Both could share the same contract.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/NoOpFilter.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.postprocessing
+
+import org.apache.commons.configuration.Configuration
+import za.co.absa.spline.harvester.HarvestingContext
+import za.co.absa.spline.producer.model.v1_1.{DataOperation, ExecutionEvent, ExecutionPlan, ReadOperation, WriteOperation}
+
+class NoOpFilter extends Filter {
```
wajda: If `spline.postprocessing_filter.classNames` was optional `NoOpFilter` class would become redundant and could be removed.

core/src/test/scala/za/co/absa/commons/CaptureGroupReplacerSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+  val Replacement = "*****"
+  val replacer = new CaptureGroupReplacer(Replacement)
+
+  it should "not modify the input when there is no match" in {
+    val input = "some string"
+    replacer.replace(input, Seq("""(\d+)""".r)) should be theSameInstanceAs input
+  }
+
+  it should "replace the capturing groups" in {
+    replacer.replace("foo@bar.cz", Seq("""[^@]+@(\w+).\w+""".r)) shouldBe s"foo@${Replacement}.cz"
```
wajda, line 33: nit: redundant curly braces - `$Replacement`

core/src/test/scala/za/co/absa/spline/harvester/builder/BuilderSpec.scala
```diff
@@ -17,26 +17,23 @@
...
 import za.co.absa.spline.harvester.{ComponentCreatorFactory, HarvestingContext}
-import za.co.absa.spline.producer.model.v1_1.ReadOperation
 
 class BuilderSpec extends AnyFlatSpec with Matchers with MockitoSugar {
 
   it should "not force lowercase on keys of the params Map when" in {
     val ctxMock = mock[HarvestingContext]
     val componentCreatorFactoryMock = mock[ComponentCreatorFactory]
-    val userExtraMetadataProviderMock = mock[UserExtraMetadataProvider]
+    val noOpPostProcessor = new PostProcessor(Seq(new NoOpFilter), ctxMock)
```
wajda: In unit-testing we should avoid involving real dependencies in the process, and use test doubles instead.

core/src/test/scala/za/co/absa/spline/harvester/postprocessing/DataSourceUriFilterSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    val filter = new DataSourceUriFilter()
+    val pp = new PostProcessor(Seq(filter), mock[HarvestingContext])
+
+    val filteredOp = pp.process(wop)
+
+    filteredOp.outputSource shouldEqual uri("*****")
+  }
+
+}
```
wajda: nit: missing empty line

core/src/test/scala/za/co/absa/spline/harvester/postprocessing/DataSourceUriFilterSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    val wop = WriteOperation(List.empty, uri("123456"), false, "42", List.empty, None, None)
+
+    val filter = new DataSourceUriFilter()
+    val pp = new PostProcessor(Seq(filter), mock[HarvestingContext])
+
+    val filteredOp = pp.process(wop)
+
+    filteredOp.outputSource shouldEqual uri("*****")
+  }
```
wajda: Usually in unit-testing, DAMP principle should be preferred over DRY. 
Trivial constant and code repetition leads to more readable and descriptive code, which considering unit-tests purposes is beneficial.
Consider this variant. It's longer but arguably more readable
```suggestion
  it should "apply one filter" in {
    val wop = WriteOperation(
      Nil,
      "" +
        "jdbc:sqlserver://database.windows.net:1433" +
        ";user=sample" +
        ";password=123456" +
        ";encrypt=true" +
        ";trustServerCertificate=false" +
        ";hostNameInCertificate=*.database.windows.net" +
        ";loginTimeout=30:",
      append = false, "", Nil, None, None)

    val filter = new DataSourceUriFilter()
    val pp = new PostProcessor(Seq(filter), mock[HarvestingContext])
    val filteredOp = pp.process(wop)

    filteredOp.outputSource shouldEqual "" +
      "jdbc:sqlserver://database.windows.net:1433" +
      ";user=sample" +
      ";password=*****" + // <-- PASSWORD SHOULD BE SANITIZED
      ";encrypt=true" +
      ";trustServerCertificate=false" +
      ";hostNameInCertificate=*.database.windows.net" +
      ";loginTimeout=30:"
  }
```

wajda: Also notice that that test method tests two classes at once: `DataSourceUriFilter` and `PostProcessor`. 
A unit test should (almost) always only test _one_ class and _one_ method at a time.
(Perhaps the only exception from this rule could be a situation when you test a pair of mutually related classes, like read/write, get/set etc, testing which in complete isolation is often difficult)

core/src/test/scala/za/co/absa/spline/harvester/postprocessing/PostProcessorSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+    override def processWriteOperation(op: WriteOperation, ctx: HarvestingContext): WriteOperation =
+      op.copy(outputSource = op.outputSource + str)
+  }
+
+  val wop = WriteOperation(List.empty, "foo", false, "42", List.empty, None, None)
+
+  it should "apply one filter" in {
+
+    val filter = new UriAppendingFilter("@")
+    val pp = new PostProcessor(Seq(filter), mock[HarvestingContext])
```
wajda: Same as above. Please use mocks and fakes to isolate `PostProcessor` from a concrete `Filter` implementation.

core/src/test/scala/za/co/absa/commons/CaptureGroupReplacerSpec.scala
cerveada: This is on purpose. I find it easier to see the variable. Github will not highlight it so it's easy to miss and takes some effort to find what part of string is a variable. The braces make it easy.

```Scala
s"foo@$Replacement.cz"

s"foo@${Replacement}.cz"
```

core/src/test/scala/za/co/absa/spline/harvester/postprocessing/PostProcessorSpec.scala
cerveada: UriAppendingFilter is a mock

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/Filter.scala
cerveada: I changed the name to `LineageFilter`

PR #141, opened 2020-10-19
Capture session config

liangjun-jiang:
This is related to feature request: https://github.com/AbsaOSS/spline-spark-agent/issues/140
In this way, we can see and use configuration set by Spark app developers.

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
@@ -97,6 +97,7 @@ class LineageHarvester(
       val plan = {
         val planExtra = Map[String, Any](
           ExecutionPlanExtra.AppName -> ctx.session.sparkContext.appName,
+          ExecutionPlanExtra.Config -> ctx.session.sparkContext.getConf.getAll,
           ExecutionPlanExtra.DataTypes -> componentCreatorFactory.dataTypeConverter.values,
           ExecutionPlanExtra.Attributes -> componentCreatorFactory.attributeConverter.values
         )
--- a/core/src/main/scala/za/co/absa/spline/harvester/ModelConstants.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/ModelConstants.scala
@@ -25,6 +25,7 @@ object ModelConstants {
 
   object ExecutionPlanExtra {
     val AppName = "appName"
+    var Config = "config"
     val DataTypes = "dataTypes"
     val Attributes = "attributes"
   }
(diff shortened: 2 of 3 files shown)


wajda:
Thank you Liangjun,
Could you add some integration tests please?

P.S. ignore failing auto builds for now. There is a problem with the CI env, we're working on the fix.

liangjun-jiang:
I don't know that much about Scala. Would like to give it a try to make some integration test cases.

wajda:
Thanks. Take a look at the existing test cases, for example `BasicIntegrationTests`, to get an idea of how it can be written.
Instead of `withSparkSession` you can use `withCustomSparkSession` method. It gives you an access to a `SparkSession.Builder` instance which you can use to set test config parameters, and later check if they appear in the resulted execution plan.

Also please note that among the session config params there might be ones logically related to the execution plan (the ones that are rather stable, e.g. `"spark.app.name"`), and the others that carry dynamic information associated with the given execution event. In Spline execution plan and event and connected as _one-to-many_. Even though current implementation of the agent produce both entities per every application run, in future it could change. (Think for example about streaming job - one execution plan with many microbatch events).
So when capturing session config params it's better to try to (as long as it's possible) logically split them, putting the stable values into the _execution plan_ and the dynamic value to the `execution event` entity.

liangjun-jiang:
I think there is no extra integration test needed. I look at this [line 344 of LIneageHarvesterSpec](https://github.com/AbsaOSS/spline-spark-agent/blob/develop/integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala#L343), it basically supports any custom tag in `extraInfo`

wajda:
No, that unit-test tests that user extra data is recorded in the `extraInfo` section.
But here we need to test how the session config is captured and serialized. If all data is serializable? What kind of data is captured? If there is not too unnecessary noise being captured, and so forth. A unit test is not enough, we need a proper integration test.

wajda:
Sorry, the test you pointed is an integration test, not a unit one, my mistake. By my comments still hold. That test is focused on a slightly different thing.

liangjun-jiang:
I took a look at this possible change. It's much beyond my Scala skill. The configuration implementation is basically very minimal.

wajda:
After reviewing it for the second time I think we should close it as this functionality is basically achievable by using `za.co.absa.spline.harvester.extra.UserExtraMetaDataProvider`. See https://github.com/AbsaOSS/spline-spark-agent/issues/35#issuecomment-629865953
Also #169 will introduce a generic mechanism of doing the same thing in a declarative way.

PR #153, opened 2020-11-24
Spark 3.0 support

shanebell:
I've added a new bundle-3.0 for Spark 3 support and made a few minor tweaks.

This compiles ok and the basics seem to work when running against a simple Spark 3.0 job using PySpark. I can see the execution event in the Spline UI and explore it etc and it all looks ok. However, the integration tests are failing.

I had a look at some of these and I'm a bit out of my depth. I'm not very familiar with Scala so I'm fumbling around a bit.

If you have any bandwidth to have a quick look at this and provide and guidance I'd appreciate your feedback.

Thanks.
Shane.


pom.xml
```diff
@@ -508,10 +509,6 @@
                 <scala.binary.version>2.11</scala.binary.version>
                 <scala.version>${scala_2.11.version}</scala.version>
             </properties>
-            <modules>
-                <module>bundle-2.2</module>
-                <module>bundle-2.3</module>
-            </modules>
```
wajda, line 514: Selection bundle modules for building should be done in Scala profiles. There could only be one Scala version per Maven run, while Spark version is selected per module (in respective poms). So one Maven run builds all bundles that support given Scala version. Roughly saying `spark-x.y` profile only tells Maven what Spark version to run integration tests with, but it doesn't affect the final artifacts.

This is what probably breaks the build for certain scala+spark combinations.

wajda:
Hi @shanebell 
What's the status of this PR. Are you still working on this?
Thanks.

shanebell:
Hi, sorry for the lack of activity on this. I'm on holiday at the moment but definitely plan to come back to this.

wajda:
Awesome! Have a nice holiday! :smile:

cerveada:
Replaced by PR #192

PR #155, opened 2020-11-27
Feature/agent 61 producer api upgrade

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/ComponentCreatorFactory.scala
```diff
@@ -18,21 +18,19 @@ package za.co.absa.spline.harvester
...
-  val expressionConverter = new ExpressionConverter(dataConverter, dataTypeConverter, attributeConverter)
+  val inputAttributeConverter = new InputAttributeConverter(dataTypeConverter)
+  val expressionConverter = new GenericExpressionConverter(dataConverter, dataTypeConverter)
   val operationParamsConverter = new OperationParamsConverter(dataConverter, expressionConverter)
 
   private[this] val lastId = new AtomicInteger(0)
 
   def nextId: Int = lastId.getAndIncrement()
+
+  def expressionConverters():List[ExpressionConverter] = inputAttributeConverter :: expressionConverter :: Nil
```
wajda:
```suggestion
  def expressionConverters():List[ExpressionConverter] = Seq(inputAttributeConverter, expressionConverter)
```

Operator `::` while still being useful for extracting and pattern-matching, as a way of composing lists however is considered deprecated since about Scala 2.8.
Respective `apply()` functions or `:+`, `:+` etc methods are usually more readable.

core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
```diff
@@ -122,7 +130,7 @@ class LineageHarvester(
         ).optionally[Duration]((m, d) => m + (ExecutionEventExtra.DurationNs -> d.toNanos), result.toOption)
 
         val ev = ExecutionEvent(
-          planId = plan.id,
+          planId = plan.id.get,
```
wajda: As an exec plan ID is generated a few lines above can we reuse the value directly instead of unwrapping an `Option` here?
```suggestion
          planId = planId,
```

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -35,13 +34,22 @@ trait OperationNodeBuilder {
...
+
+  protected def childIds: List[OperationId] = childBuilders.map(_.id).toList
+  protected def childOutputSchemas: Seq[OutputAttIds] = childBuilders.map(_.outputAttributes)
   protected def isTerminal: Boolean = childBuilders.isEmpty
 }
 
 object OperationNodeBuilder {
-  type Schema = Seq[UUID]
+  type OperationId = String
+  type OutputAttIds = List[String]
```
wajda, line 53: Why `List`, not `Seq` or array?

core/src/main/scala/za/co/absa/spline/harvester/builder/GenericNodeBuilder.scala
```diff
@@ -35,9 +35,9 @@ class GenericNodeBuilder
   override def build(): DataOperation = {
     val dop = DataOperation(
       id = id,
-      childIds = childIds.toList.asOption,
-      schema = if (!isTerminal && childOutputSchemas.forall(outputSchema.==)) None else Some(outputSchema),
```
wajda, line 39: The schema used to be optional, isn't it such anymore?

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
```diff
@@ -18,98 +18,53 @@ package za.co.absa.spline.harvester.converter
...
-        getExpressionExtraParameters(e))
+import org.apache.spark.sql.catalyst.expressions.ExprId
+import za.co.absa.spline.harvester.builder.OperationNodeBuilder.OperationId
+import za.co.absa.spline.harvester.converter.ExpressionConverter.ExpressionLike
+import za.co.absa.spline.model.TempReference
+import za.co.absa.spline.producer.model.v1_1.{Attribute, FunctionalExpression, Literal}
+
+import scala.collection.mutable
+
+trait ExpressionConverter {
```
wajda: Why trait, not an abstract class?

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -35,13 +34,22 @@ trait OperationNodeBuilder {
...
 
   protected def componentCreatorFactory: ComponentCreatorFactory
-  protected lazy val outputSchema: Schema = operation.output.map(componentCreatorFactory.attributeConverter.convert(_).id)
-
-  protected def childIds: Seq[Int] = childBuilders.map(_.id)
-  protected def childOutputSchemas: Seq[Schema] = childBuilders.map(_.outputSchema)
+  protected lazy val outputAttributes: OutputAttIds = {
+    val convert: Attribute => String =
+      if (isTerminal)
+        att => componentCreatorFactory.inputAttributeConverter.convert(att, id).id
```
wajda: Can we extract a converter (as function of `isTerminal`) outside of this block, so that we can reuse this piece better without having branching inside the map body?

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
```diff
@@ -18,98 +18,53 @@ package za.co.absa.spline.harvester.converter
...
+  def convert(sparkExpr: expressions.Expression, operationId: String): ExpressionLike
+
+  private val attributeStorage = mutable.Buffer[Attribute]()
+  private val literalStorage = mutable.Buffer[Literal]()
+  private val functionalExpressionStorage = mutable.Buffer[FunctionalExpression]()
```
wajda: Is it necessary to use mutable collections here? If an imperative algorithm is desired can we use immutable `var` instead of mutable `val`?

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
```diff
@@ -18,98 +18,53 @@ package za.co.absa.spline.harvester.converter
...
   }
-
-  private def getDataType(expr: SparkExpression) = dataTypeConverter.convert(expr.dataType, expr.nullable)
 }
 
 object ExpressionConverter {
-  private val basicProperties = Set("children", "dataType", "nullable")
 
-  private def getExpressionSimpleClassName(expr: SparkExpression) = {
+  type ExpressionLike = {
```
wajda, line 34: This type alias is possibly redundant. 
From I see it's mainly used as a return value from the `ExpressionConverter.convert()` method. In all the invocation places it's only used to get `id` from it immediately after the method call, so just returning an ID seems to be enough. The method itself doesn't seem to use this type either. So the role of `ExpressionLike` is unclear.

core/src/main/scala/za/co/absa/spline/harvester/converter/GenericExpressionConverter.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  override def convert(sparkExpr: SparkExpression, operationId: String): ExpressionLike = {
+    val output = convertInternal(sparkExpr, operationId)
+    store(output)
+    output
+  }
+
+  private def convertInternal(sparkExpr: SparkExpression, operationId: String): ExpressionLike = sparkExpr match {
+
+  case a: expressions.Alias =>
```
wajda: nit: indentation

core/src/main/scala/za/co/absa/spline/harvester/converter/InputAttributeConverter.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Expression, Attribute => SparkAttribute}
+import za.co.absa.spline.harvester.builder.OperationNodeBuilder.OperationId
+import za.co.absa.spline.harvester.converter.ExpressionConverter.{ExpressionLike, toSplineAttrId, createExtra}
+import za.co.absa.spline.producer.model.v1_1.Attribute
+
+class InputAttributeConverter(dataTypeConverter: DataTypeConverter) extends ExpressionConverter {
+
+  def convert(sparkExpr: Expression, operationId: String): ExpressionLike = {
+    val output = convertInternal(sparkExpr, operationId)
+    store(output)
```
wajda: This looks like a use-case for a decorator pattern to me. Also what about SRP? Is the class responsible for converting, or collecting/persisting the value?

core/src/test/scala/za/co/absa/spline/harvester/converter/ExpressionConverterSpec.scala
```diff
@@ -37,30 +39,29 @@ class ExpressionConverterSpec extends AnyFlatSpec with OneInstancePerTest with M
...
 
   when(dtConverterMock convert NullType -> true) thenReturn nullDataType
   when(dtConverterMock convert StringType -> false) thenReturn stringDataType
 
   it should "support secondary constructor, but only capture params from the primary one" in {
     val expression = new Foo("this parameter should not be captured")
 
-    inside(converter convert expression) {
-      case expr.Generic(name, dataTypeId, children, exprType, Some(params)) =>
+    inside(converter convert(expression, opId)) {
```
wajda: code style: infix notation

core/src/test/scala/za/co/absa/spline/harvester/converter/ExpressionConverterSpec.scala
```diff
@@ -37,30 +39,29 @@ class ExpressionConverterSpec extends AnyFlatSpec with OneInstancePerTest with M
...
     }
   }
 
   it should "ignore properties with null or None value" in {
     val expression = Foo.empty
 
-    inside(converter convert expression) {
-      case expr.Generic(_, _, _, _, Some(params)) =>
+    inside(converter convert(expression, opId)) {
+      case FunctionalExpression(_, _, _, _, _, params) =>
```
wajda: The number of `_` is growing. Consider replacing an extraction pattern with a type matching pattern followed by an accessor call.

core/src/main/scala/za/co/absa/spline/harvester/converter/OperationParamsConverter.scala
```diff
@@ -24,25 +24,28 @@ import org.apache.spark.sql.types.DataType
...
+  override type From = (LogicalPlan, String)
   override type To = Map[String, _]
 
-  private val renderer = ValueDecomposer.addHandler(_ => {
+  private def renderer(operationId: String) = ValueDecomposer.addHandler(_ => {
     case (row: InternalRow, rowType: DataType) => Some(dataConverter.convert((row, rowType)))
     case (jt: JoinType, _) => Some(jt.sql)
     case (so: SortOrder, _) => Some(Map(
-      "expression" -> expressionConverter.convert(so.child),
+      "expression" -> new ExpressionReference(expressionConverter.convert(so.child, operationId).id),
```
wajda: `ExpressionReference` doesn't have a state. Can we avoid `new` then?

core/src/main/scala/za/co/absa/spline/model/TempReference.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.model
+
+case class TempReference(id: String)
```
wajda: The class is very specific to the implementation of expression converters. Can it be moved to a narrower scope of visibility? Also a sentence of scala doc would help to understand it's role.

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
cerveada: producer model uses lists, so I used the list here as well

core/src/main/scala/za/co/absa/spline/harvester/builder/GenericNodeBuilder.scala
cerveada: It is not optional. 

We had expressions in the op.params. Now the output is the main connection between operation and attribute. So I think it makes sense to have the schema everywhere.

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
cerveada: After the code review changes, the Convertor API with output storing convertor as is used. To store the Expressions I have to return them, so using just the Id is not viable any more, I also need to get the id on many places. `ExpressionLike` is the solution for this.

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionStoringConverter.scala
```diff
@@ -0,0 +1,45 @@
...
+
+trait ExpressionStoringConverter extends ExpressionConverter {
+
+  private var attributeStorage = List.empty[Attribute]
+  private var literalStorage = List.empty[Literal]
+  private var functionalExpressionStorage = List.empty[FunctionalExpression]
+
+  def attributes: List[Attribute] = attributeStorage.reverse
+  def literals: List[Literal] = literalStorage.reverse
+  def functionalExpressions: List[FunctionalExpression] = functionalExpressionStorage.reverse
```
wajda: The return type can changed to `Seq` with no changes required in other places.

core/src/main/scala/za/co/absa/spline/harvester/ComponentCreatorFactory.scala
```diff
@@ -18,21 +18,19 @@ package za.co.absa.spline.harvester
...
-  val expressionConverter = new ExpressionConverter(dataConverter, dataTypeConverter, attributeConverter)
+  val inputAttributeConverter = new InputAttributeConverter(dataTypeConverter) with ExpressionStoringConverter
+  val expressionConverter = new GenericExpressionConverter(dataConverter, dataTypeConverter) with ExpressionStoringConverter
   val operationParamsConverter = new OperationParamsConverter(dataConverter, expressionConverter)
 
   private[this] val lastId = new AtomicInteger(0)
 
   def nextId: Int = lastId.getAndIncrement()
+
+  def expressionConverters(): List[ExpressionStoringConverter] = List(inputAttributeConverter, expressionConverter)
```
wajda: Minor: Since this method is idempotent, it has better be defined as ether a parameterless method or a value.

wajda (commented):
Just two small things, otherwise LGTM.

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -35,13 +34,21 @@ trait OperationNodeBuilder {
...
+      componentCreatorFactory.inputAttributeConverter
+    else
+      componentCreatorFactory.expressionConverter
+
+  private def convert(att: SparAttribute): String = attributeConverter.convert((att, id)).id
+  protected lazy val outputAttributes: OutputAttIds = operation.output.map(convert).toList
 
-  protected def childIds: Seq[Int] = childBuilders.map(_.id)
-  protected def childOutputSchemas: Seq[Schema] = childBuilders.map(_.outputSchema)
+  protected def childIds: List[OperationId] = childBuilders.map(_.id).toList
```
wajda: Here we can use `Seq` as well with no extra efforts. The `.toList` call will just move from here to the place of usage.

Issue #158, opened 2020-12-11
REST Dispatcher: wrong Content-Type for v1.1 API


wajda:
Should be `application/vnd.absa.spline.producer.v1.1+json`

PR #159, opened 2020-12-11
spline #158 REST Dispatcher: wrong Content-Type for v1.1 API

wajda:


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -18,82 +18,93 @@ package za.co.absa.spline.harvester.dispatcher
...
-  private def createHttpErrorMessage(msg: String, code: Int, body: String): String =
-    s"$msg. HTTP Response: $code $body"
+  override def send(event: ExecutionEvent): Unit = {
+    val eventDTO = modelMapper.toDTO(event)
+    sendJson(Seq(eventDTO).toJson, executionEventsEndpoint)
+  }
+
+  private def sendJson(json: String, endpoint: RestEndpoint) = {
+    val url = endpoint.request.url
+    logTrace(s"sendJson $url : $json")
```
cerveada: Since we are logging this, is there an easy way to format the json before it is logged?

wajda: parse and re-render with pretty-print probably. Will do

wajda: updated

cerveada (approved):

Issue #65, opened 2020-05-10
Attribute level lineage isn't resolved correctly on UNION operation


wajda:
Even though UNION's inputs have identical schema the output attributes are only traced to the 1st input group.

Expected behavior: UNION output should be traced to both input groups.

![image.png](https://images.zenhubusercontent.com/5c04fd55820d8748e06666cd/b4754545-3527-467a-bdd1-eb2530a5e180)

PR #163, opened 2021-01-19
#65 union expression level lineage

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/converter/OperationParamsConverter.scala
```diff
@@ -24,27 +24,25 @@ import org.apache.spark.sql.types.DataType
...
+  override type From = LogicalPlan
   override type To = Map[String, _]
 
-  private def renderer(operationId: String) = ValueDecomposer.addHandler(_ => {
+  private def valueDecomposer = ValueDecomposer.addHandler(_ => {
     case (row: InternalRow, rowType: DataType) => Some(dataConverter.convert((row, rowType)))
     case (jt: JoinType, _) => Some(jt.sql)
     case (so: SortOrder, _) => Some(Map(
-      "expression" -> ExpressionReference(expressionConverter.convert((so.child, operationId)).id),
+      "expression" -> expressionConverter.toAttrOrExprRef(expressionConverter.convert(so.child)),
```
wajda, line 39: `expressionConverter.toAttrOrExprRef` vs `expressionConverter.convert`
Suggestions:
 - use move the `toAttrOrExprRef` to a separate class and use that instead of `ExpressionConverter`, or
- replace `toAttrOrExprRef` method with `ExpressionLik.reference` member method

core/src/main/scala/za/co/absa/spline/harvester/converter/InputAttributeConverter.scala
```diff
@@ -16,24 +16,22 @@
...
 
-import org.apache.spark.sql.catalyst.expressions.{Expression, Attribute => SparkAttribute}
-import za.co.absa.spline.harvester.builder.OperationNodeBuilder.OperationId
-import za.co.absa.spline.harvester.converter.ExpressionConverter.{ExpressionLike, toSplineAttrId, createExtra}
+import org.apache.spark.sql.catalyst.expressions.{Attribute => SparkAttribute, Expression => SparkExpression}
+import za.co.absa.spline.harvester.converter.ExpressionConverter.{ExpressionLike, createExtra}
 import za.co.absa.spline.producer.model.v1_1.Attribute
 
-class InputAttributeConverter(dataTypeConverter: DataTypeConverter) extends ExpressionConverter {
+class InputAttributeConverter(
```
wajda, line 23: From the methods' signatures this class supposes to convert _any_ Spark expression into Spline `ExpressionLike`, but in fact in can only process Spark attributes.

wajda: Since we only concern about _input_ attributes (as the class name suggests). why not to use `AttributeReference` instead of an abstract 'Attribute'?

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionAccumulator.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.converter
+
+import za.co.absa.spline.harvester.converter.ExpressionConverter.{ExpressionLike, TempReference}
+import za.co.absa.spline.producer.model.v1_1.{Attribute, FunctionalExpression, Literal}
+
+class ExpressionAccumulator {
```
wajda, line 22: Why do we need this? Don't `CachingConverter`'s capabilities suffice?

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -40,15 +40,14 @@ trait OperationNodeBuilder {
     else
       componentCreatorFactory.expressionConverter
 
-  private def convert(att: SparAttribute): String = attributeConverter.convert((att, id)).id
+  private def convert(att: SparAttribute): String = attributeConverter.convert(att).id
   protected lazy val outputAttributes: OutputAttIds = operation.output.map(convert).toList
```
wajda, line 44: redundant `toList`

core/src/main/scala/za/co/absa/spline/harvester/builder/GenericNodeBuilder.scala
```diff
@@ -31,14 +31,22 @@ class GenericNodeBuilder
   override protected type R = DataOperation
 
   override def build(): DataOperation = {
+
+    enrich(operation)
+
     val dop = DataOperation(
       id = id,
       childIds = childIds.toList.asOption,
```
wajda, line 39: redundant `toList`

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -31,7 +31,7 @@ trait OperationNodeBuilder {
 
   def operation: LogicalPlan
   def build(): R
-  def +=(childBuilder: OperationNodeBuilder): Unit = childBuilders :+= childBuilder
+  def addChild(childBuilder: OperationNodeBuilder): Unit = childBuilders :+= childBuilder
 
   protected def componentCreatorFactory: ComponentCreatorFactory
   private lazy val attributeConverter =
```
wajda, line 37: Terminal operations don't have any input attributes, but `inputAttributeConverter` is selected. This doesn't seem right.

core/src/main/scala/za/co/absa/spline/harvester/builder/GenericNodeBuilder.scala
```diff
@@ -31,14 +31,22 @@ class GenericNodeBuilder
...
       output = outputAttributes,
-      params = componentCreatorFactory.operationParamsConverter.convert((operation, id)).asOption,
+      params = componentCreatorFactory.operationParamsConverter.convert(operation).asOption,
       extra = Map(OperationExtras.Name -> operation.nodeName).asOption
     )
 
     postProcessor.process(dop)
   }
+
+  private def enrich(logicalPlan: LogicalPlan): Unit = logicalPlan match {
```
wajda, line 48: What is this method for?
The verb _enrich_ suggests that some extra data will be added to the passed object (either in a mutable or immutable way). The return type `Unit` tells that in should be _mutable_ addition. However the implementation doesn't do anything of above. Instead, it calls another external method for a pure side-effect.

core/src/main/scala/za/co/absa/spline/harvester/converter/UnionLineageCreator.scala
```diff
@@ -0,0 +1,82 @@
...
+package za.co.absa.spline.harvester.converter
+
+import org.apache.spark.sql.catalyst.expressions.{Attribute => SparkAttribute}
+import org.apache.spark.sql.catalyst.plans.logical.Union
+import za.co.absa.commons.lang.OptionImplicits.NonOptionWrapper
+import za.co.absa.spline.producer.model.v1_1.{AttrOrExprRef, Attribute, FunctionalExpression, Literal}
+
+import java.util.UUID
+
+class UnionLineageCreator(dataTypeConverter: DataTypeConverter, attributeIdResolver: AttributeIdResolver){
```
wajda, line 26: What does this _creator_ create?
The only public method declares a return type `Any` while in fact always returning `Unit`?

core/src/main/scala/za/co/absa/spline/harvester/converter/GenericExpressionConverter.scala
```diff
@@ -17,63 +17,62 @@
...
-        childIds = List(convert(a.child, operationId).id),
-        extra = createExtra(a, "expr.Attribute", operationId),
+        childIds = List(toAttrOrExprRef(convert(a.child))).asOption,
+        extra = createExtra(a, "expr.Attribute"),
         name = a.name
       )
 
     case a: expressions.Attribute =>
-      TempReference(toSplineAttrId(a.exprId))
+      TempReference(attributeIdResolver.resolve(a.exprId))
```
wajda, line 48: What is the purpose of `TempReference`? The only place where it is actually used in the code is this:
```scala
case ar: TempReference => AttrOrExprRef(Some(ar.id), None)
```
Can't we directly use `AttrOrExprRef` type instead?

core/src/main/scala/za/co/absa/spline/harvester/converter/AttributeIdResolver.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.converter
+
+import org.apache.spark.sql.catalyst.expressions.ExprId
+
+import scala.collection.mutable
+
+class AttributeIdResolver {
```
wajda, line 23: Is this class actually needed? All it does is wrapping a mutable map. The usage of the composite key also seems to be unnecessary, as we a priori are dealing with a logical execution plan created on the same JVM.

core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
```diff
@@ -142,37 +142,45 @@ class LineageHarvester(
     })
   }
 
+  private trait StackItem
```
wajda, line 145: As these private types have no dependency on the `LigeageHarvester` instance, them had better be put in the respective companion object. Also think about moving the build creation logic into another class.

core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
```diff
@@ -142,37 +142,45 @@ class LineageHarvester(
...
+  private case class NewItem(plan: LogicalPlan, parentBuilder: OperationNodeBuilder) extends StackItem
+  private case class ProcessedItem(builder: OperationNodeBuilder) extends StackItem
+
   private def createOperationBuildersRecursively(rootOp: LogicalPlan): Seq[OperationNodeBuilder] = {
```
wajda, line 149: Does this method return builders in a topological order? If so, what piece of code is responsible for the node precedence?

core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionAccumulator.scala
cerveada: They do not, 
 - it's available only for converters
 - I don't want to do caching at all, every object will be converted just once - no point in using cache
 - Id is not available before the conversion, part of the conversion is to get the id

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
cerveada: Terminal operations define the input attributes of the whole ExecutionPlan. From the POV of the whole plan it makes sense.

core/src/main/scala/za/co/absa/spline/harvester/converter/InputAttributeConverter.scala
cerveada: > Since we only concern about _input_ attributes (as the class name suggests). why not to use `AttributeReference` instead of an abstract 'Attribute'?

Why not program to interface/abstract class? Why make it more specific than necessary?

PR #167, opened 2021-02-01
Feature/spline spark agent 165

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
```diff
@@ -42,9 +42,8 @@ class QueryExecutionEventHandler(
       .harvest(Try(durationNs.nanos))
       .foreach({
         case (plan, event) =>
-          val idAsJson = lineageDispatcher.send(plan)
-          val savedPlanId = UUID.fromString(idAsJson.fromJson[String])
-          lineageDispatcher.send(event.copy(planId = savedPlanId))
+          val id = lineageDispatcher.send(plan)
+          lineageDispatcher.send(event.copy(planId = id))
```
wajda: If I'm not mistaken we've come into conclusion that ID aliasing should be implemented on the server (_if_ is implemented at all), and the server would always respond with the same ID (or a new one if one wasn't provided).
But since we know that the agent always generates IDs the `event.copy(planId = id)` expression becomes redundant.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -94,35 +96,55 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
...
-  protected def lineageDispatcher: LineageDispatcher = instantiate[LineageDispatcher](
-    configuration.getRequiredString(LineageDispatcherClass))
-
-  protected def ignoredWriteDetectionStrategy: IgnoredWriteDetectionStrategy = instantiate[IgnoredWriteDetectionStrategy](
-    configuration.getRequiredString(IgnoreWriteDetectionStrategyClass))
+  protected def lineageDispatcher: LineageDispatcher =
+    configuration
+      .getOptionalString(DeprecatedConfProperty.LineageDispatcherClass)
+      .map(instantiate[LineageDispatcher])
+      .getOrElse{
```
wajda: Shouldn't it be in the opposite way - read the new property first, and if one isn't found then try the deprecated one?
Instead of dealing with deprecated properties directly in the same place where actual properties are dealt with, I'd suggest to try to isolate them in the dedicated `Configuration` adapter, and then attach it to the configuration chain right before the `default.properties`.

E.g.
```scala
private lazy val configuration = new CompositeConfiguration(Seq(
    userConfiguration,
    new Spline05ConfigurationAdapter(userConfiguration), <--- This one here!
    new PropertiesConfiguration(defaultPropertiesFileName)
  ).asJava)
```
`Spline05ConfigurationAdapter` would translate the new properties to the old ones and vice versa.
This way the original config stack would be re-tried for the old properties, and if not found the `default.properties` would be used for the new ones.

The backward compatibility logic would be 100% encapsulated in a dedicated class, better testable, leaving the rest of config logic clean.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -94,35 +95,45 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
   override def queryExecutionEventHandler: QueryExecutionEventHandler =
     new QueryExecutionEventHandler(harvesterFactory, lineageDispatcher)
 
-  protected def lineageDispatcher: LineageDispatcher = instantiate[LineageDispatcher](
-    configuration.getRequiredString(LineageDispatcherClass))
+  protected def lineageDispatcher: LineageDispatcher = {
+    val namespace = configuration.getRequiredString(LineageDispatcherNamespace)
```
wajda: Minor: here `namespace` variable actually represents a _logical dispatcher name_, so I'd suggest to rename it respectively.

core/src/main/scala/za/co/absa/spline/harvester/conf/Spline05ConfigurationAdapter.scala
```diff
@@ -0,0 +1,64 @@
...
+import java.util
+import scala.collection.JavaConverters.asJavaIterableConverter
+
+class Spline05ConfigurationAdapter(configuration: Configuration) extends ReadOnlyConfiguration {
+
+  override def isEmpty: Boolean = scalaKeys.isEmpty
+
+  override def containsKey(key: String): Boolean = keyMap.get(key).map(configuration.containsKey).getOrElse(false)
+
+  override def getProperty(key: String): AnyRef = keyMap.get(key).map(configuration.getProperty).getOrElse(null)
```
wajda: When is called with `spline.lineageDispatcher` as a key, it's supposed to return a default dispatcher name (in this case it would be `http`). But it should return it if and only if the `spline.lineage_dispatcher.className` is specified. That is because the "old" entry `spline.lineage_dispatcher.className=FooBar` is represented by two "new" entries: `spline.lineageDispatcher=something` + `spline.lienageDispatcher.something.className=FooBar`. So in reverse, if the config source reports that `spline.lienageDispatcher.something.className` is defined, then by definition the complimentary property `spline.lienageDispatcher` with the value.`something` has to also be defined.

wajda: I think this class wants some unit tests.

core/src/main/scala/za/co/absa/spline/harvester/conf/Spline05ConfigurationAdapter.scala
```diff
@@ -0,0 +1,74 @@
...
+  override def containsKey(key: String): Boolean = keyMap
+    .get(key)
+    .map(spline05key => configuration.containsKey(spline05key))
+    .getOrElse(defaultValues.contains(key))
+
+  override def getProperty(key: String): AnyRef = keyMap
+    .get(key)
+    .map(spline05key => configuration.getProperty(spline05key))
+    .orElse(defaultValues.get(key))
+    .getOrElse(null)
```
wajda: nit: `.getOrElse(null)` is the same as `.orNull`

core/src/test/scala/za/co/absa/spline/harvester/conf/Spline05ConfigurationAdapterTest.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+}
+
+object Spline05ConfigurationAdapterTest {
+
+  private def createConfigAdapter(map: Map[String, Any]) =
+    new Spline05ConfigurationAdapter(new MapConfiguration(map.asJava))
+
+  val testConfig = createConfigAdapter(Map(
```
wajda: nit: + `private`

wajda (dismissed):
LGTM excepts for two minor comments.

Issue #172, opened 2021-02-16
Composite lineage dispatcher


wajda:
Create a `LineageDispatcher` implementation that delegate the lineage data to one or more other dispatchers.

wajda:
@Zejnilovic @benedeki

PR #173, opened 2021-02-17
spline #172 Composite lineage dispatcher

wajda:


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/CompositeLineageDispatcher.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package za.co.absa.spline.harvester.dispatcher
+
+import za.co.absa.commons.config.ConfigurationImplicits._
+import za.co.absa.spline.harvester.conf.HierarchicalObjectFactory
+import za.co.absa.spline.harvester.dispatcher.CompositeLineageDispatcher._
+import za.co.absa.spline.producer.model.v1_1.{ExecutionEvent, ExecutionPlan}
+
+class CompositeLineageDispatcher(delegatees: LineageDispatcher*) extends LineageDispatcher {
```
cerveada:
```suggestion
class CompositeLineageDispatcher(delegates: LineageDispatcher*) extends LineageDispatcher {
```

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/CompositeLineageDispatcher.scala
```diff
@@ -0,0 +1,54 @@
...
+  override def send(plan: ExecutionPlan): Unit = delegate {
+    _.send(plan)
+  }
+
+  override def send(event: ExecutionEvent): Unit = delegate {
+    _.send(event)
+  }
+
+  private def delegate(call: LineageDispatcher => Unit): Unit = {
+    delegatees.foreach(call)
```
cerveada: Shouldn't there be some exception handling, so exception in one dispatcher doesn't stop the other?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/CompositeLineageDispatcher.scala
wajda: No, it's delegatees :) https://en.wiktionary.org/wiki/delegatee

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/CompositeLineageDispatcher.scala
wajda: added a property `failOnErrors` which if set to `true` re-throws exceptions and fails fast,
or otherwise suppresses the errors. logging them as WARN and proceeds. (Default is `false` - suppressing mode)

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/CompositeLineageDispatcher.scala
cerveada: ok

cerveada (approved):

Issue #166, opened 2021-01-26
Kafka lineage dispatcher


wajda:
Lineage info to be optionally sent over the Kafka.
Execution plans and events to be sent in different topics.

Config example:
```properties
# "kafka" dispatcher config
spline.dispatcher.kafka.className=za.co.absa.spline.harvester.dispatcher.HttpLineageDispatcher
spline.dispatcher.kafka.topic.execPlans=
spline.dispatcher.kafka.topic.execEvents=
```

PR #179, opened 2021-02-25
agent #166 add kafka lineage dispatcher

cerveada:


core/src/main/resources/spline.default.properties
```diff
@@ -31,6 +31,16 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
...
 
+# Kafka dispatcher
+spline.lineageDispatcher.kafka.className=za.co.absa.spline.harvester.dispatcher.KafkaLineageDispatcher
+# producer configs as defined by kafka (bootstrap.servers, key.serializer, etc) all kafka configs are supported
+spline.lineageDispatcher.kafka.producer.bootstrap.servers=
+spline.lineageDispatcher.kafka.producer.key.serializer=org.apache.kafka.common.serialization.StringSerializer
+spline.lineageDispatcher.kafka.producer.value.serializer=org.apache.kafka.common.serialization.StringSerializer
+# topic names for plan and event
+spline.lineageDispatcher.kafka.planTopic=spline-plans
+spline.lineageDispatcher.kafka.eventTopic=spline-events
```
wajda: These two don't have any natural defaults, do they? Let's leave them empty. Let's make the user to specify them in order to use this dispatcher. (The same I would do on the server side). The motivation is that even though it does reduce the configuration burden on the quick starting, but might create a pitfall in some real-life usages. Same as with the preset database names for instance. Even though it's unlikely, if one connected it to a shared Kafka cluster to have say Dev and UAT topics, one must also not forget to change the default values. And if it's unintentionally left by default, the packages would be routed to a wrong topic and it would be hard to spot and correct this mistake. (I remember how in one well known reputable international company where I previously worked, a devops connected a UAT application server to the Prod database before calling it a day and leaving the office. That was fun. So these things do happen :)

core/src/main/scala/za/co/absa/commons/Charest.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.commons
+
+import java.nio.charset.{Charset => JavaCharset}
+
+object Charset {
+  val `UTF-8` = JavaCharset.forName("UTF-8")
```
wajda: There is already a constant in JDK - ` java.nio.charset.StandardCharsets.UTF_8`

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/kafkadispatcher/KafkaLineageDispatcherConfig.scala
```diff
@@ -0,0 +1,38 @@
...
+object KafkaLineageDispatcherConfig {
+  val planTopicKey = "planTopic"
+  val eventTopicKey = "eventTopic"
+  val producerKey = "producer"
+
+  def apply(c: Configuration) = new KafkaLineageDispatcherConfig(c)
+}
+
+class KafkaLineageDispatcherConfig(config: Configuration) {
+  val planTopic = config.getRequiredString(planTopicKey)
```
wajda: Public members should have defined types

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/KafkaLineageDispatcher.scala
```diff
@@ -0,0 +1,84 @@
...
+import za.co.absa.spline.harvester.json.HarvesterJsonSerDe.impl._
+import za.co.absa.spline.producer.model.v1_1.{ExecutionEvent, ExecutionPlan}
+
+import scala.collection.JavaConverters.asJavaIterableConverter
+import scala.util.control.NonFatal
+
+/**
+ * KafkaLineageDispatcher is responsible for sending the lineage data to spline gateway through kafka
+ */
+class KafkaLineageDispatcher(config: KafkaLineageDispatcherConfig)
```
wajda: It's Ok, but just wondering why have you introduced a _parameter object_ here? There are just three independent arguments. The other dispatchers define theirs params directly in the constructor and it looks sufficient and straightforward to me. Do you have another opinion or you plan adding more parameters to the class?

cerveada: I thought there will be more properties, but since producer have its own subconfig it ended up being only two for now.

I will remove it.

wajda (approved):
LGTM

PR #180, opened 2021-02-28
Move binary expression "symbol" property from params to extras

wajda:
The "params" are supposed to contain stuff that affects the expression output in any way (e.g. tuning parameters, number formats etc). The "symbol" being a static representational property (an alias to the "name") fits extras better.

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
@@ -54,9 +54,9 @@ class ExpressionConverter(
         id = randomUUIDString,
         dataType = convertDataType(bo),
         childIds = convertChildren(bo).asOption,
-        extra = createExtra(bo, "expr.Binary").asOption,
+        extra = (createExtra(bo, "expr.Binary") + ("symbol" -> bo.symbol)).asOption,
         name = bo.prettyName,
-        params = (getExpressionParameters(bo) + ("symbol" -> bo.symbol)).asOption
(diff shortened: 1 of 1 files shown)


cerveada:
> The "params" are supposed to contain stuff that affects the expression output in any way (e.g. tuning parameters, number formats etc). The "symbol" being a static representational property (an alias to the "name") fits extras better.

That is new information for me.

I thought `params` are any parameters and `extra` is user defined info (Using Filters, ExtraMetadataPrivder, etc) that also contains some values for backward compatibility.

Even the API documentation seems to suggest this.
```
  // [Optional] Named expression parameters
  params: {...},
  // [Optional] Custom meta info
  extra: {...}
```

wajda:
That's correct. From the Producer API perspective _expression_ is a function that is applied on the set of values (that is a function input) and might also have optional other parameters in addition to the input. The _extra_ is a model extension point, it might contain any meta info that an agent (or a user) wants to keep linked to the expression for whatever reason.

In other word - the "params" have a relation to the function semantics, while "extras" do not.

cerveada (approved):

wajda:
Using analogy of e.g. Scala `foldLeft` function.
```scala
nums.foldLeft(0)(_+_) // a sum
nums.foldLeft(1)(_*_) // a factorial
```
It the above example `nums` would be a function input, while an accumulator value (0 or 1) would be a function parameter. An accumulator affects the output, but isn't presented as a part of input collection.
The name of function and its arity can be seen as extra meta data, that describes the function or the context of the invocation, but isn't variable for the given function and doesn't logically affect its behavior.

PR #184, opened 2021-03-16
Examples: + pseudo DCE examples

wajda:


examples/src/main/scala/za/co/absa/spline/example/dce/MyOtherJobC.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  spark.enableLineageTracking()
+
+  spark.read
+    .csv("data/output/dce/publish/Transactions/2020/01/01/v1")
+    .write
+    .mode("overwrite")
+    .save("data/_temp/my-other-job-output")
+
+}
```
cerveada: Could the code duplication be refactored away? 
Finding out if the Jobs are really identical takes some effort. If they used the same function (for example) it would be much easier to be sure.

cerveada (dismissed):
I would remove the code duplication, otherwise LGTM

examples/src/main/scala/za/co/absa/spline/example/dce/MyOtherJobC.scala
wajda: I tend to be conservative with code reuse in tests and examples. They are made primarily for reading and copy/pasting by users that are looking at the project for the 1st time. Any loop, inheritance or even a helper method call would not contribute to readability in comparison with a few-line of flat plain and pure Spark code snippet. I would even inline "extends SparkApp", but that's OK. As long as you can copy/past it into the spark-shell and it works, it's fine.

cerveada: As a 1st time reader I had a problem to understand whether MyOtherJobs are very similar or identical. So that doesn't contribute to readability either.

I agree that it makes sense to keep things simple. But these examples make sense only when run together. And therefore I have to understand what they are supposed to do together and for that It really helps to know that the code is identical. Which is easy to see when there is just one code.

Alternatively there could be just one Job and you could just run it three times.

cerveada:
Seems like the duplicated Jobs (A, B, C) won't be run by Maven. Is that intended?
```
<target name="runAll">
...
  <include name="**/example/**/*Job.class"/>
...
</target>
```

wajda:
> Seems like the duplicated Jobs (A, B, C) won't be run by Maven. Is that intended?
> 
> ```
> <target name="runAll">
> ...
>   <include name="**/example/**/*Job.class"/>
> ...
> </target>
> ```

No, it's not intended :) Thx.

PR #186, opened 2021-03-17
#185 - windows build fix. 

### LINKED ISSUE #185: java.lang.RuntimeException: missing OpenAPI input! when tryin to build the package
Hi,

When trying to build the agent  I am hitting the following error:
```
[INFO] --- swagger-codegen-maven-plugin:3.0.23:generate (swagger.api.v1_0) @ agent-core_2.12 ---
[WARNING] Output directory does not exist, or is inaccessible. No file (.swagger-codegen-ignore) will be evaluated.
[ERROR]
java.lang.RuntimeException: missing OpenAPI input!
    at io.swagger.codegen.v3.DefaultGenerator.generate (DefaultGenerator.java:768)
```
Anyone has encountered this/has any ideas?

Thanks a lot

dk1844:
- Mainly, downgrading swagger codegen to 3.0.21 from 3.0.23 (src: https://www.gitmemory.com/issue/swagger-api/swagger-codegen/10844/764741128)
 - other small fields generation fixes (usage did not match the generated case classes)

Closes #185

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -134,7 +134,7 @@
             <plugin>
                 <groupId>io.swagger.codegen.v3</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
-                <version>3.0.21</version>
+                <version>3.0.21</version> <!-- please note 3.0.22+ were problematic on Windows. See Issue #185 -->
                 <configuration>
                     <language>scala</language>
                     <generateModels>true</generateModels>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -134,7 +134,7 @@
             <plugin>
                 <groupId>io.swagger.codegen.v3</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
-                <version>3.0.23</version>
+                <version>3.0.21</version>
                 <configuration>
                     <language>scala</language>
                     <generateModels>true</generateModels>
--- a/core/src/main/scala/za/co/absa/spline/harvester/converter/AttributeConverter.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/converter/AttributeConverter.scala
@@ -40,6 +40,7 @@ class AttributeConverter(
         dataType = Some(dataTypeConverter.convert(attr.dataType, attr.nullable).id),
         childIds = resolveAttributeChild(attr)
           .map(expr => Seq(exprToRefConverter.convert(expr))),
+        extra = None,
         name = attr.name
       )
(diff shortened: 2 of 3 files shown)


cerveada:
The fields you added are None by default for me:

```
package za.co.absa.spline.producer.model.v1_1

case class Attribute (
  id: String,
  dataType: Option[Any] = None,
  childIds: Option[Seq[AttrOrExprRef]] = None,
  extra: Option[Map[String, Any]] = None,
  name: String
)

// in another file:

case class ExecutionEvent (
  planId: UUID,
  timestamp: Long,
  error: Option[Any] = None,
  extra: Option[Map[String, Any]] = None
)
```

Are the generated classes different on Windows?

---

I built this branch on a Mac and the missing defaults seems to be caused by the version not the OS,
because they are not there even for me.

core/pom.xml
```diff
@@ -134,7 +134,7 @@
             <plugin>
                 <groupId>io.swagger.codegen.v3</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
-                <version>3.0.23</version>
+                <version>3.0.21</version>
```
wajda: Could you please add a comment here with a reference to the swagger plugin issue? Just to warn the other dev from upgrading to a more recent version (for whatever reason) until the aforementioned issue is fixed.

dk1844: Sure.

dk1844:
> The fields you added are None by default for me:
> 
> ```
> package za.co.absa.spline.producer.model.v1_1
> 
> case class Attribute (
>   id: String,
>   dataType: Option[Any] = None,
>   childIds: Option[Seq[AttrOrExprRef]] = None,
>   extra: Option[Map[String, Any]] = None,
>   name: String
> )
> 
> // in another file:
> 
> case class ExecutionEvent (
>   planId: UUID,
>   timestamp: Long,
>   error: Option[Any] = None,
>   extra: Option[Map[String, Any]] = None
> )
> ```
> 
> ## Are the generated classes different on Windows?
> I built this branch on a Mac and the missing defaults seems to be caused by the version not the OS,
> because they are not there even for me.

Yes, the default values are not generated for me on Windows with Swagger v3.0.21, hence their addition.
defaults-not-generated

wajda:
> The fields you added are None by default for me:

It seems defaults aren't generated by the older swagger plugin, that's why those changes were needed.

wajda (approved):
LGTM

cerveada (approved):
LGTM

PR #192, opened 2021-03-26
Feature/spline-93 - Spark 3.0 support

cerveada:


bundle-3.0/pom.xml
```diff
@@ -0,0 +1,1329 @@
...
+                <groupId>com.github.luben</groupId>
+                <artifactId>zstd-jni</artifactId>
+                <scope>provided</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <build>
+        <plugins>
+            <plugin>
```
wajda, line 1265: The _develop_ branch contains "mima" plugin that needs to be disabled in bundle poms.
Add this piece here:
```xml
            <plugin>
                <groupId>org.spurint.maven.plugins</groupId>
                <artifactId>mima-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <id>check-abi</id>
                        <phase>none</phase>
                    </execution>
                </executions>
            </plugin>
```

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/KafkaPlugin.scala
```diff
@@ -41,11 +42,25 @@ class KafkaPlugin
...
       val options = extractFieldValue[Map[String, String]](kr, "sourceOptions")
-      val topics: Seq[String] = extractFieldValue[ConsumerStrategy](kr, "strategy") match {
-        case AssignStrategy(partitions) => partitions.map(_.topic)
-        case SubscribeStrategy(topics) => topics
-        case SubscribePatternStrategy(pattern) => kafkaTopics(options("kafka.bootstrap.servers")).filter(_.matches(pattern))
-      }
+      // org.apache.spark.sql.kafka010.ConsumerStrategy - private classes cannot be used directly
+      val consumerStrategy = extractFieldValue[AnyRef](kr, "strategy")
+
+      def tryAssignStrategy(): Try[Seq[String]] =
```
wajda: nit: redundant parenthesis

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/KafkaPlugin.scala
```diff
@@ -41,11 +42,25 @@ class KafkaPlugin
...
+        Try(extractFieldValue[Seq[String]](consumerStrategy, "topics"))
+
+      def trySubscribePatternStrategy(): Try[Seq[String]] =
+        Try(extractFieldValue[String](consumerStrategy, "topicPattern"))
+          .map(pattern => kafkaTopics(options("kafka.bootstrap.servers")).filter(_.matches(pattern)))
+
+      val topics: Seq[String] =
+        tryAssignStrategy
+          .getOrElse(trySubscribeStrategy
+            .getOrElse(trySubscribePatternStrategy.get))
```
wajda:
```suggestion
        tryAssignStrategy
          .orElse(trySubscribeStrategy)
          .orElse(trySubscribePatternStrategy)
          .get
```

... or even ... 
```suggestion
        Seq(
          tryAssignStrategy,
          trySubscribeStrategy,
          trySubscribePatternStrategy
        )
          .reduce(_ orElse _)
          .get
```

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/NoOpLineageDispatcher.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.dispatcher
+
+import org.apache.commons.configuration.Configuration
+import za.co.absa.spline.producer.model.v1_1.ExecutionEvent
+import za.co.absa.spline.producer.model.v1_1.ExecutionPlan
+
+class NoOpLineageDispatcher extends LineageDispatcher {
```
wajda: Why do we need it?

bundle-3.0/pom.xml
```diff
@@ -0,0 +1,1329 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>spark-3.0-spline-agent-bundle_2.11</artifactId>
```
wajda: such artifact does not exist. The suffix should be `2.12` for this module.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -124,3 +126,22 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
       .getOrElse(postProcessingFilters)
   )
 }
+
+case class CodeBasedSplineConfigurer(
```
wajda: Is this class used? (I guess it's unfinished due to the draft PR status)

bundle-3.0/pom.xml
cerveada: `scala-cross-build:restore-version` always changes it to `2.11`. Is there some way to override the default Scala version just for this bundle?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/NoOpLineageDispatcher.scala
cerveada: Used for testing user extra metadata, but it could probably be moved to a test folder.

bundle-3.0/pom.xml
wajda: hm.. it shouldn't. This module is only connected to Maven when you activate `scala-2.12` profile. So whatever plugin you execute, unless `scala-2.12` profile is activated this POM should stay disconnected and hence unaffected.

wajda: try to remove all `pom.xml.backup` and try again.

cerveada: `scala-cross-build:restore-version -P scala-2.12` sets the suffix to _2.11 in spark 3 bundle. No matter what was there on the start. 

`scala-cross-build:restore-version -P scala-2.11` doesn't change spark 3 bundle.

I don't think the backups are used at all. I think the plugin just takes the default version and overwrite the suffixies by it.

wajda: Why would you call `restore` with a specific profile? To me the command only makes sense when is called alone. The case when a set of modules is different depending on which Scala version profile you choose doesn't seem to be covered by the plugin. So I'd say simply don't call `restore` and instead always use `change-version`.
The main point here is that the `pom.xml` should by default be in the valid state, i.e. spark-3.0_2.12`

core/src/main/scala/za/co/absa/commons/FutureImplicits.scala
```diff
@@ -0,0 +1,48 @@
...
+      * Executes the body after the future is completed (no matter if successfully or not)
+      * Returns the future with the same value/exception
+      *
+      * Analogous to finally block in try-catch-finally statement
+      * Unlike onComplete this guarantees that body will be executed before any other operations on the future
+      *
+      * @param body function executed for a side effect
+      * @return unchanged future that it was called on
+      */
+    def finallyDo(body: => Unit)(implicit executor: ExecutionContext): Future[T] = {
```
wajda: Why not using `Future.andThen()`?

bundle-3.0/pom.xml
cerveada: It would be nice to have a command that would restore everything to default version, so I can erase any changes due to version switching. (for example before committing changes) But that doesn't seem to be possible right now.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/NoOpLineageDispatcher.scala
cerveada: Moved to test package/folder

wajda (approved):

PR #193, opened 2021-03-26
HDFSLineageDispatcher uses runtime json4s dependencies

dk1844:
...in order to prevent failing on binary incompatibilities (e.g. between json4s 3.5.3 and 3.5.5)

Tested successfully in Enceladus.

Thanks @wajda for hints on this.


wajda:
This issue is going to reoccur in other embedded dispatchers that extend `AbstractJsonLineageDispatcher`
Can you please expand the fix so that it covers those ones as well?

Think about reusing `HarvesterJsonSerDe` for this purpose.

@dk1844 replying: sure; //edit: done.

core/src/test/scala/za/co/absa/spline/harvester/dispatcher/ConsoleLineageDispatcherSpec.scala
```diff
@@ -29,20 +31,29 @@ class ConsoleLineageDispatcherSpec
...
   behavior of "ConsoleLineageDispatcher"
 
-  it should "send lineage to the stdout" in {
-    assertingStdOut(include("""["event",{"planId":null,"timestamp":999}]""")) {
-      new ConsoleLineageDispatcher(new BaseConfiguration {
-        addProperty("stream", "OUT")
-      }).send(ExecutionEvent(null, 999, None, None))
+  private val uuid1 = UUID.randomUUID()
+
+  Seq(
```
wajda: Using loops, branching or other conditional logic is generally not recommended practice in unit-tests.
~DRY~ DAMP principle works great in test code - the simpler (flatter) the better.
If you really need to share a test code (rather rarely), there are better ways for test code sharing (e.g. property-style testing, or behavior-style testing).
In this case I'd say there is nothing to reuse, so simply adding additional "should" statements would suffice.

wajda (commented):
LGTM, except for the test case styling.

core/src/test/scala/za/co/absa/spline/harvester/dispatcher/ConsoleLineageDispatcherSpec.scala
dk1844: Sure, your castle (project), your rules. I have removed the branching, now there are only flat should-statements.

(But still, I had the feeling that this actually fulfilled the DRY principle as there was less repeating. I see, however, the probable reasoning behind keeping the tests rather simple - both for readability and not to actually recreate the logic in `main`).

wajda: Did I say DRY? That was a typo, I apologize. I actually meant DAMP :)

wajda (approved):
LGTM

PR #195, opened 2021-03-29
S3 lineage support

dk1844:
This change introduces S3 support for the "HDFS Lineage Dispatcher". 

~The naming change (HDFS -> HadoopFs in the class name and also `hdfs` -> `hadoopFs` in the config space) is meant to communicate that not only classic HDFS is supported, but other filesystems, too, if accessed via the Hadoop FS API (e.g. S3 for example).~

Tested in custom Enceladus: https://github.com/AbsaOSS/enceladus/pull/1742


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HadoopFsLineageDispatcher.scala
```diff
@@ -88,16 +97,38 @@ class HDFSLineageDispatcher(filename: String, permission: FsPermission, bufferSi
...
+        (fs, s3Path)
+
+      case None => // local hdfs location
+        val fs = FileSystem.get(HadoopConfiguration)
+        (fs, new Path(pathString))
+    }
   }
-}
+
+}
```
cerveada: We are striving to have a line break at the end of each file.

Idea can do this for you:
Screenshot 2021-03-29 at 14 34 03

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.commons
+
+import scala.util.matching.Regex
+
+object S3Location {
```
wajda, line 20: Why is it in the `commons` package?
Unless you are planing (or anticipating) to use it for different things in the agent core, and unless it's  generic enough to be moved to the "Commons" later, it should rather be placed close to the class that uses it, e.g. `za.co.absa.....dispatcher.hfs.S3Localtion`

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Generally usable regex for validating S3 path, e.g. `s3://my-cool-bucket1/path/to/file/on/s3.txt`
+   * Protocols `s3`, `s3n`, and `s3a` are allowed.
+   * Bucket naming rules defined at [[https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules]] are instilled.
+   */
+  val S3LocationRx: Regex = "(s3[an]?)://([-a-z0-9.]{3,63})/(.*)".r
+
+  implicit class StringS3LocationExt(path: String) {
```
wajda: consider making it a value-class

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.util.matching.Regex
+
+object S3Location {
+
+  /**
+   * Generally usable regex for validating S3 path, e.g. `s3://my-cool-bucket1/path/to/file/on/s3.txt`
+   * Protocols `s3`, `s3n`, and `s3a` are allowed.
+   * Bucket naming rules defined at [[https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules]] are instilled.
+   */
+  val S3LocationRx: Regex = "(s3[an]?)://([-a-z0-9.]{3,63})/(.*)".r
```
wajda: Shouldn't this constant be private?

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+    def toS3LocationOrFail: SimpleS3Location = {
+      this.toS3Location.getOrElse{
+        throw new IllegalArgumentException(s"Could not parse S3 Location from $path using rx $S3LocationRx.")
+      }
+    }
+
+    def isValidS3Path: Boolean = path match {
+      case S3LocationRx(_, _, _) => true
+      case _ => false
+    }
```
wajda:
```suggestion
    def isValidS3Path: Boolean = S3LocationRx.pattern.matcher(s).matches()
```

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+  val S3LocationRx: Regex = "(s3[an]?)://([-a-z0-9.]{3,63})/(.*)".r
+
+  implicit class StringS3LocationExt(path: String) {
+
+    def toS3Location: Option[SimpleS3Location] = {
+      path match {
+        case S3LocationRx(protocol, bucketName, relativePath) => Some(SimpleS3Location(protocol, bucketName, relativePath))
+        case _ => None
+      }
+    }
```
wajda:
```suggestion
    def toS3Location: Option[SimpleS3Location] = PartialFunction.condOpt(s) {
      case S3LocationRx(protocol, bucketName, relativePath) => 
        SimpleS3Location(protocol, bucketName, relativePath)
    }
```

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+
+    def toS3LocationOrFail: SimpleS3Location = {
```
wajda: Why do you need both `toS3Location` and `toS3LocationOrFail` methods?
If it's only about different code styles then it could be worth it if it were some kind or DSL implementation, but is it really worth it in this case? (Likewise you wouldn't add extra methods here returning `Try` or `Either` that only differs in return type, would you?)

What about only leaving the one that returns `Option`? The caller could always switch to any other alternatives in the place of invocation.

core/src/main/scala/za/co/absa/commons/S3Location.scala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    def isValidS3Path: Boolean = path match {
+      case S3LocationRx(_, _, _) => true
+      case _ => false
+    }
+  }
+}
+
+trait S3Location {
```
wajda: Why do we need this trait?

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: I have a very similar thing in Atum, so, yes, moving this eventually to _the_ or _a_ commons package would make sense to me. Do you think that _the_ commons would be a good place for it?

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: You are absolutely right, I am (as outline above) aiming for a general-purpose usage. The extra stuff here, we don't need, will remove.

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: Good idea, will do.

core/src/main/scala/za/co/absa/commons/S3Location.scala
wajda: Definitely, especially if you are already using it for more than one project, that's what Commons exists for.

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: Sure. Done.

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: Thanks, suggestion taken.

core/src/main/scala/za/co/absa/commons/S3Location.scala
dk1844: Not needed as part of this solution, perhaps later for the general S3Location solution. removing.

core/src/main/scala/za/co/absa/commons/S3Location.scala
wajda: seems like compilation fails due to the typo in the suggested code

wajda:
There is a failed test, can you check it please? https://teamcity.jetbrains.com/buildConfiguration/OpenSourceProjects_AbsaOSS_SplineAgentSpark_AutoBuildSpark23scala211/3402624?buildTab=tests&name=HadoopFsLineageDispatcherSpec&expandedTest=-6361448792504931515

@dk1844 replying: yes, checking it out. It seems that the _LINEAGE file is not found, I will investigate.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HadoopFsLineageDispatcher.scala
dk1844: Thanks for the tip, @cerveada! Done.

dk1844:
> There is a failed test, can you check it please? https://teamcity.jetbrains.com/buildConfiguration/OpenSourceProjects_AbsaOSS_SplineAgentSpark_AutoBuildSpark23scala211/3402624?buildTab=tests&name=HadoopFsLineageDispatcherSpec&expandedTest=-6361448792504931515
> 
> @dk1844 replying: yes, checking it out. It seems that the _LINEAGE file is not found, I will investigate.

To my surprise, the problem occurs only on CI and only when renaming `HDFSLineageDispatcher`->`HadoopFsLineageDispatcher` (and related) and `hdfs` ->`hadoopfs` is in place. I couldn't find what makes the `hdfs`-variant special, so I have gone back to the original naming that works without problem on the CI.

On TC CI, the `hadoopfs` className is not respected and the default `http` dispatcher is used instead, of course failing.
```log
[08:18:24][Step 2/3] DEBUG za.co.absa.spline.harvester.conf.HierarchicalObjectFactory  - Instantiating za.co.absa.spline.harvester.dispatcher.HttpLineageDispatcher
[08:18:24][Step 2/3] ERROR za.co.absa.spline.harvester.QueryExecutionEventHandlerFactory  - Spline initialization failed! Spark Lineage tracking is DISABLED.
[08:18:24][Step 2/3] java.util.NoSuchElementException: Missing configuration property spline.lineageDispatcher.http.producer.url
```

The same behaves correctly on local:
```log
DEBUG  za.co.absa.spline.harvester.conf.HierarchicalObjectFactory  - Instantiating za.co.absa.spline.harvester.dispatcher.HadoopFsLineageDispatcher
INFO  za.co.absa.spline.harvester.QueryExecutionEventHandlerFactory  - Spline successfully initialized. Spark Lineage tracking is ENABLED.
```

cerveada (approved):

PR #198, opened 2021-04-01
Update build-all.sh

wajda:
Do a proper clean up after the cross build.
Extract a global list of supported Scala binary versions.

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -17,6 +17,8 @@
 #
 # THIS SCRIPT IS INTENDED FOR LOCAL DEV USAGE ONLY
 #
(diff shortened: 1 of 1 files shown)


build-all.sh
```diff
@@ -18,21 +18,41 @@
 # THIS SCRIPT IS INTENDED FOR LOCAL DEV USAGE ONLY
 #
```
cerveada, line 19:
```suggestion
#
# Build Spline Agent artifacts for all supported Scala versions and install them to local maven repository.
#
```

Short comment explaining what the script does might be useful

cerveada (dismissed):

cerveada (approved):

PR #203, opened 2021-04-09
Swagger defs update

wajda:
According to AbsaOSS/spline#831


cerveada (approved):
LGTM, but I would wait for Gateway PR before merging this one.

Issue #200, opened 2021-04-09
"null" should be a valid literal


wajda:
The agent is now using the removing-empty-values serialization strategy that doesn't assume existence of semantically valid `null` values, e.g. in literals. That results in producing invalid JSON that misses required properties (e.g. literal value that happens to be `null`)

```
com.twitter.finatra.json.internal.caseclass.exceptions.CaseClassValidationException: expressions.constants.value: field is required
```
 
We need to fix serialization in a way that allows for having `value: null` where appropriate.

wajda:
@dk1844, @AdrianOlosutean

wajda:
the simplest and quickest fix would be adding a default value to the producer model case class on the server side, but that would result in effectively making `value` an optional property for literals which is logically incorrect. So, to preserve the current level of Producer API model correctness I'd rather fix it on the agent.

wajda:
Seems that `NoEmptyValueSupport` isn't needed in the current agent version as the majority of the meaningless empty values used to come from the reflected entities, but the currently used revision of `ValueDecomposer` seems to take care of it naturally during recursion. Running example jobs without `NoEmptyValueSupport` revealed no impact, so I guess the one can be safely removed and the original issue will be fixed.

PR #204, opened 2021-04-15
spline #200 Remove `NoEmptyValueSupport`

wajda:
Fixes missing `null` values in required places.
This trait seems to be redundant anyway

### DIFF
--- a/core/src/test/scala/za/co/absa/spline/harvester/dispatcher/ConsoleLineageDispatcherSpec.scala
+++ b/core/src/test/scala/za/co/absa/spline/harvester/dispatcher/ConsoleLineageDispatcherSpec.scala
@@ -33,15 +33,7 @@ class ConsoleLineageDispatcherSpec
 
   private val uuid1 = UUID.fromString("12345678-90ab-cdef-1234-567890abcdef")
 
-  it should "send lineage to the stdout (null planId)" in {
-    assertingStdOut(include("""["event",{"timestamp":999}]""")) {
-      new ConsoleLineageDispatcher(new BaseConfiguration {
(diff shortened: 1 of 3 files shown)


cerveada (dismissed):

core/src/test/scala/za/co/absa/spline/harvester/dispatcher/ConsoleLineageDispatcherSpec.scala
```diff
@@ -33,31 +33,15 @@ class ConsoleLineageDispatcherSpec
...
 
-  it should "send lineage to the stderr (null planId)" in {
-    assertingStdErr(include("""["event",{"timestamp":999}]""")) {
-      new ConsoleLineageDispatcher(new BaseConfiguration {
-        addProperty("stream", "ERR")
-      }).send(ExecutionEvent(null, 999, None, None, None))
-    }
-  }
-
-  it should "send lineage to the stderr (non-null planId)" in {
```
dk1844, line 60: Why is it no longer desirable to check that the json serialization result with the null fields has the expected form?

wajda: That test was irrelevant anyway. The serialization logic should be tested on the serializer level. I'll add a test (thought I added one, but apparently I removed it during cleanup before committing)

wajda: added. please check

cerveada (approved):

Issue #201, opened 2021-04-09
Missing attribute data when using producer 1.0 protocol


wajda:
# Test case:
Agent 0.6.0 --> Server 0.5.6

PR #208, opened 2021-04-23
spline #201 Missing attribute data when using producer 1.0 protocol

wajda:
Fixed model v1 to v1.1 converter


wajda:
manual tests passed for the server ver 0.4.2, 0.5.6 as well as the latest one.

image

cerveada (approved):
LGTM, but some unit test for the mapper might be useful

wajda:
> LGTM, but some unit test for the mapper might be useful

I wish we had more time for that :(

PR #220, opened 2021-05-07
bugfix/spline-69 hide more passwords

wajda:
- Scan for sensitive read/write operation param names and values
- make default regexes case-insensitive
- change the filter logical name from `dsUriPasswordReplace` to `dsPasswordReplace`

### DIFF
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -101,5 +101,5 @@ spline.postProcessingFilter.dsPasswordReplace.nameRegexes=\
   (?i)password ,\
   (?i)passphrase
 spline.postProcessingFilter.dsPasswordReplace.valueRegexes=\
-  (?i)[?&;]password=([^&;]*)(?\=[&;])? ,\
+  (?i)[?&;](?:password|passphrase)=([^&;]*)(?\=[&;])? ,\
   //[^:]*:([^@]*)@(?:\\w+\\.)*\\w+
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -90,13 +90,16 @@ spline.postProcessingFilter.composite.className=za.co.absa.spline.harvester.post
 # The classes must have a constructor with a single parameter of type {org.apache.commons.configuration.Configuration}
 #
 # Example:
-#   to enable password replacing in the data source URIs add "dsUriPasswordReplace" filter to the list:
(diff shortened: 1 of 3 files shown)


core/src/main/scala/za/co/absa/spline/harvester/postprocessing/DataSourcePasswordReplacingFilter.scala
```diff
@@ -20,32 +20,56 @@ import org.apache.commons.configuration.Configuration
...
+      outputSource = filter(op.outputSource),
+      params = op.params.map(filter)
+    )
 
-  private val replacer = new CaptureGroupReplacer(replacement)
+  private val valueReplacer = new CaptureGroupReplacer(replacement)
 
   private def filter(uri: String): String =
-    replacer.replace(uri, regexes)
+    valueReplacer.replace(uri, sensitiveValueRegexes)
```
cerveada: It's not necessary uri anymore. Maybe `uri` -> `value`

core/src/main/resources/spline.default.properties
```diff
@@ -90,13 +90,16 @@ spline.postProcessingFilter.composite.className=za.co.absa.spline.harvester.post
...
-spline.postProcessingFilter.dsUriPasswordReplace.replacement=*****
-spline.postProcessingFilter.dsUriPasswordReplace.regexes=\
-  [?&;]password=([^&;]*)(?\=[&;])? ,\
+spline.postProcessingFilter.dsPasswordReplace.className=za.co.absa.spline.harvester.postprocessing.DataSourcePasswordReplacingFilter
+spline.postProcessingFilter.dsPasswordReplace.replacement=*****
+spline.postProcessingFilter.dsPasswordReplace.nameRegexes=\
+  (?i)password ,\
+  (?i)passphrase
+spline.postProcessingFilter.dsPasswordReplace.valueRegexes=\
+  (?i)[?&;](?:password|passphrase)=([^&;]*)(?\=[&;])? ,\
```
cerveada: What exactly is `?\=` dooing in `(?\=[&;])?` ? 

Also, this regexp is not covered by the test.

wajda: It's a _lookahead_ to assert the following symbol without consuming it. But you're right, it seems unnecessary in this particular expression, as the same delimiter character is captured in the beginning of the next matching iteration anyway.

It is covered by the first test.

cerveada (approved):

PR #222, opened 2021-05-10
spline #spline-spark-agent-221 Rename synthetic attributes of a union…

wajda:
… operation to match original attribute name

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/UnionNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/UnionNodeBuilder.scala
@@ -58,8 +58,7 @@ class UnionNodeBuilder
     )
 
   private def constructUnionAttribute(attributes: Seq[Attribute], function: FunctionalExpression) = {
-    val attr1 +: attrs = attributes
-    assume(attrs.forall(_.name == attr1.name))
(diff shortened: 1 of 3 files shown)


core/src/main/scala/za/co/absa/spline/harvester/builder/UnionNodeBuilder.scala
```diff
@@ -35,34 +36,49 @@ class UnionNodeBuilder
...
-      params = Map("name" -> "union").asOption
+      extra = Map(ExtraFields.Synthetic -> true).asOption,
+      name = Names.Union,
+      params = None
     )
 
-  private def constructUnionAttribute(inputSplineAttributes: Seq[Attribute], function: FunctionalExpression) =
+  private def constructUnionAttribute(attributes: Seq[Attribute], function: FunctionalExpression) = {
+    val attr1 +: attrs = attributes
+    assume(attrs.forall(_.name == attr1.name))
```
cerveada: This assumption is wrong. Spark doesn't require the name to be the same. 

Spark docs for UNION:
> Note that input relations must have the same number of columns and compatible data types for the respective columns.

cerveada (changes requested):

cerveada (approved):

PR #229, opened 2021-06-03
Fix PostProcessingFilters custom filter name section

singhrajk:
This is to ensure accuracy of configuration examples for reference. I stumbled upon this for a while. Only realised the className here is not required.


wajda (approved):
Nice one! :smile: Thank you!

PR #234, opened 2021-06-04
Docker image tag consistency in example/

bensenberner:
The two commands are meant to be run one after the other, so they should have the same tag


wajda (approved):
Thanks. But I actually think this paragraph needs to be rewritten, as since the version 0.6.0 we publish a docker image with the agent and examples to a Docker hub - https://hub.docker.com/r/absaoss/spline-spark-agent

PR #260, opened 2021-06-22
Spline spark agent #226 cyclic annotation bug workaround

cerveada:
Workaround for Scala bug #12190

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/converter/ReflectiveExtractor.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/converter/ReflectiveExtractor.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


core/src/main/scala/za/co/absa/spline/harvester/converter/ReflectiveExtractor.scala
```diff
@@ -0,0 +1,57 @@
...
+import java.lang.reflect.{Field, Method}
+
+object ReflectiveExtractor extends Logging{
+
+  def extractProperties(obj: AnyRef): Map[String, _] =
+    try ReflectionUtils.extractProperties(obj)
+    catch {
+      // a workaround for Scala bug #12190
+      // catching everything because the bug causes various errors and exceptions
+      case e =>
```
wajda: Note that this will catch everything including fatal errors and exceptions that require special handling. To filter them out, `case NonFatal(e)` or `Try` should be used instead.

cerveada: I thought NonFatal filters out all exceptions. Including the ones cause by the Scala bug. But I looked it up, and it's only a few serious exceptions, and the rest goes through. In that case, NonFatal can be used here.

wajda (approved):
lgtm

PR #271, opened 2021-07-26
Feature/agent 191 data source v2 support

cerveada:
I was able to do most of the things in backward compatible way:
- The plugin for Delta, Cassandra and Json (probably will work for other file based DS)
- tests for delta and json

Cassandra connector is not backward compatible, so the test is not included. I don't know if it makes sense to make a special module just for that. We could just add the test to the 1.0 where it will not be a problem.


wajda (approved):
LGTM

Issue #266, opened 2021-07-14
DataSourceV2 byName attribute order


cerveada:
### Problem
Spark has two methods how to match a dataframe columns to columns of the output.
- by order (first column in df is first in table, second is second etc) (`insertInto`)
- by name (the column "a" in df corresponds to column "a" in table "b" corresponds to "b" etc) (`saveAsTable`)

The by name is potentially problematic for us because Spline does not store the column metadata or write operation output. The only thing we store is the output of operation before write.

In DataSourceV1 there is no problem because Spark creates additional projection that change the attribute order before the write.

In DataSourceV2 this seems to be a property of the actual write. There is `isByName` that tells which method is used and the output of the table has correct ordering, but the operation before write has attributes in the original order. Therefore, the ordering we have will be not correct, which is a problem for end to end attribute lineage.

### Possible Solution
- Add output to write operations and store the correct ordering there (more clean but requires api change)
- Add "synthetic" projection before the write (no api changes, but creates not existing operation and expresions)

### How to replicate

DataSourceV1 
```Scala
import spark.implicits._

for {
  (plan0, _) <- captor.lineageOf {
    Seq((1, 2)).toDF("i", "j").write.mode("overwrite").saveAsTable("t1")
  }
  (plan, _) <- captor.lineageOf {
    Seq((3, 4)).toDF("j", "i").write.mode("append").saveAsTable("t1")
  }
} yield {
  plan.operations.write.append should be(true)
}
```

DataSourceV2
```Scala
withCustomSparkSession(_
      .enableHiveSupport
      .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
      .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
    ) { implicit spark =>
      withLineageTracking { lineageCaptor =>
        withDatabase("testDB") {

          import spark.implicits._

          for {
            (plan, _) <- lineageCaptor.lineageOf {
              spark.sql(s"CREATE TABLE t1 (i int, j int) USING delta")
              Seq((3, 4)).toDF("j", "i").write.format("delta").mode("append").saveAsTable("t1")

              spark.sql("select * from t1").show()
            }
          } yield {
            plan.operations.write.append should be(true)
          }
        }
      }
    }
```

PR #274, opened 2021-07-29
spark agent #266 Data Source V2 write output reordering filter

cerveada:
filter fixing wrong output order in some DSV2 writes

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeOrderEnrichingFilter.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeOrderEnrichingFilter.scala
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2020 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.postprocessing
+
+import org.apache.commons.configuration.Configuration
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import za.co.absa.commons.reflect.ReflectionUtils._
+import za.co.absa.commons.reflect.extractors.SafeTypeMatchingExtractor
+import za.co.absa.spline.harvester.HarvestingContext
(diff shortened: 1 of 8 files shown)


core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeOrderEnrichingFilter.scala
```diff
@@ -0,0 +1,97 @@
...
+import za.co.absa.commons.reflect.ReflectionUtils._
+import za.co.absa.commons.reflect.extractors.SafeTypeMatchingExtractor
+import za.co.absa.spline.harvester.HarvestingContext
+import za.co.absa.spline.harvester.builder.UnionNodeBuilder.ExtraFields
+import za.co.absa.spline.harvester.plugin.embedded.DataSourceV2Plugin.`_: V2WriteCommand`
+import za.co.absa.spline.producer.model.v1_1.{DataOperation, ExecutionPlan, WriteOperation}
+
+import java.util.UUID
+
+class AttributeOrderEnrichingFilter(conf: Configuration) extends AbstractPostProcessingFilter {
```
wajda: Order _enriching_ or _enforcing_? Or may be _reordering_ filter?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeOrderEnrichingFilter.scala
```diff
@@ -0,0 +1,97 @@
...
+      val namedRelation = extractFieldValue[AnyRef](writeCommand, "table")
+      val finalOutput = extractFieldValue[Seq[Attribute]](namedRelation, "output")
+      val query = extractFieldValue[LogicalPlan](writeCommand, "query")
+
+      query.output.map(att => finalOutput.indexWhere(_.name == att.name))
+    }
+  }
+
+  private def getWriteChildOutput(plan: ExecutionPlan, writeOp: WriteOperation): Seq[String] = {
+    val writeChildId = writeOp.childIds.head
```
wajda: Calling `head` implies that there is a `tail` and thus it raises a question - why we only take `head`, what about the the rest items?

I would add some explanation:
```suggestion
    assume(writeOp.childIds.length == 1) // In Spark, a Write operation is expected to have a single child.
    val writeChildId = writeOp.childIds.head
```

or even better, using `unapply` suggests to me that we know the exact list structure, no tail is assumed, so all is clear.

```suggestion
    val Seq(writeChildId) = writeOp.childIds
```

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeOrderEnrichingFilter.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    plan.operations.other
+      .flatMap(_.find(_.id == writeChildId)).flatMap(_.output)
+      .orElse(plan.operations.reads.flatMap(_.find(_.id == writeChildId).flatMap(_.output)))
+      .get
```
wajda: Too many `flatMap`s per code line :) I would break this piece into a several variables or functions to improve readability.

wajda (approved):
LGTM

Issue #272, opened 2021-07-29
Duplicated attribute IDs


wajda:
**UPDATE**: A temporary workaround - https://github.com/absaoss/spline-spark-agent/issues/272#issuecomment-895947366

The issue was found in and causing AbsaOSS/spline#925

See JSON sample in https://github.com/AbsaOSS/spline/issues/925#issuecomment-874263960

minimal code to replicate the issue:
```Scala
package za.co.absa.spline

import org.apache.spark.SparkConf
import org.apache.spark.sql.functions.col
import org.apache.spark.sql.{DataFrame, SaveMode, SparkSession}

object SplineDuplicatedIds {

  val extraConf: Iterable[(String, String)] = List(
    ("spark.sql.queryExecutionListeners", "za.co.absa.spline.harvester.listener.SplineQueryExecutionListener"),
    ("spark.spline.lineageDispatcher", "console"),
    ("spark.spline.mode", "REQUIRED")
  )

  def main(args: Array[String]): Unit = {

    val sparkConf = new SparkConf()
      .setAll(extraConf)
      .setMaster("local[*]")

    val ss = SparkSession.builder().config(sparkConf).getOrCreate()
    import ss.implicits._

    val df = Seq(("adcde1_12938597", 162))
      .toDF("unique_id", "total_commission")

    val result: DataFrame = df
      .withColumn("commission", col("total_commission"))
      .drop("total_commission")

    val firstValidTransactions = result.withColumnRenamed("commission", "foo")

    val joined = result.join(firstValidTransactions, usingColumns = Seq("unique_id"))

    joined
      .write
      .mode(SaveMode.Overwrite)
      .option("path", "tmp/spline_test_bi_duplicates")
      .saveAsTable("spline_test_bi_duplicated")
  }
}
```
Happens on Spark 3.1, 2.4 and probably all others as well

wajda:
@kuhnen, can you share a Spark job source code that reproduces the issue please?
We could reconstruct it from the execution plan JSON, but with the source code it would be much easier. Thank you!

wajda:
Observation:

The sample JSON above contains 3 `Read` operations that read from the same data source. Each of them emits its own set of attributes (same name and type, but different IDs). Then there are 3 `Project` operations each follows one of those reads and outputs an attribute `cast(total_commission) AS commission`. For some reason all those `commission` attributes are given the same ID!! (in this example is `123`)

kuhnen:
[...]

PR #281, opened 2021-08-12
spark agent #272 fix duplicated attribute ids

cerveada:
- remove deprecated symbol `/:`

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/JoinNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/JoinNodeBuilder.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.harvester.builder
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.plans.logical.Join
+import za.co.absa.spline.harvester.ComponentCreatorFactory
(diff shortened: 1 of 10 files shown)


core/src/main/scala/za/co/absa/spline/harvester/builder/JoinNodeBuilder.scala
```diff
@@ -0,0 +1,40 @@
...
+class JoinNodeBuilder
+  (override val operation: Join)
+  (override val componentCreatorFactory: ComponentCreatorFactory, postProcessor: PostProcessor)
+  extends GenericNodeBuilder(operation)(componentCreatorFactory, postProcessor) with Logging {
+
+  override def build(): DataOperation = {
+
+    val duplicates = operation.output.groupBy(_.exprId).collect { case (x, List(_,_,_*)) => x }
+    if (duplicates.nonEmpty) {
+      logError(s"Duplicated attributes found in Join operation output, ExprIds of duplicates: $duplicates")
```
wajda: We log it as an _Error_, but there is no recovery logic and we just carry on. Shouldn't we log it as _Warning_ instead?

core/src/main/scala/za/co/absa/spline/harvester/builder/JoinNodeBuilder.scala
```diff
@@ -0,0 +1,40 @@
...
+import za.co.absa.spline.producer.model.v1_1.DataOperation
+
+class JoinNodeBuilder
+  (override val operation: Join)
+  (override val componentCreatorFactory: ComponentCreatorFactory, postProcessor: PostProcessor)
+  extends GenericNodeBuilder(operation)(componentCreatorFactory, postProcessor) with Logging {
+
+  override def build(): DataOperation = {
+
+    val duplicates = operation.output.groupBy(_.exprId).collect { case (x, List(_,_,_*)) => x }
```
wajda: The `operation.output` type is `Seq`, but the `List(_,_,_*)` would only work for _Lists_, not for other _Seqs_.
We should use a more generic type destructor `Seq(_, _, _*)`. Or even better, since we are only interested in count but not the values itself, we could communicate this logic more clearly:
```suggestion
    val duplicates = operation.output.groupBy(_.exprId).collect { case (exprId, attrs) if attrs.length > 1 => exprId }
```

wajda (approved):
LGTM

kuhnen:
Amazing @cerveada  :).  Do you guys have plans to deploy a new version of the agent?

wajda:
I'm working on it at the moment. Will trigger the release shortly after finishing some testing.

kuhnen:
amazing :). thanks a lot @wajda

PR #282, opened 2021-08-13
spark agent #262 special Databricks Window handling

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/builder/WindowNodeBuilder.scala
```diff
@@ -16,19 +16,31 @@
...
   extends GenericNodeBuilder(operation)(componentCreatorFactory, postProcessor) {
 
   override def resolveAttributeChild(attribute: Attribute): Option[Expression] = {
-    operation.windowExpressions
+    extractNamedExpressions(operation)
       .find(_.exprId == attribute.exprId)
       .map(_.asInstanceOf[Alias])
   }
+
+  private def extractNamedExpressions(operation: Window): Seq[NamedExpression] = {
```
wajda, line 41: I would add a comment explaining why just `.name` doesn't suffice, with the reference to this issue.

cerveada:
This still needs to be tested on Databricks.

wajda (approved):

wajda (approved):

Issue #290, opened 2021-08-25
Conversion to Json Error (scala.MatchError: MethodType(TypeRefType ... )


cerveada:
```
21/08/25 17:04:46 ERROR SplineQueryExecutionListener: Unexpected error occurred during lineage processing for application: DSV2Job #local-1629903864734
scala.MatchError: MethodType(TypeRefType(ThisType(scala),scala.Function1,List(TypeRefType(ThisType(scala),scala.Any,List()), TypeRefType(ThisType(scala),scala.Any,List()))),List(MethodSymbol(from, owner=45, flags=2000, info=123 ,None), MethodSymbol(to, owner=45, flags=2000, info=123 ,None))) (of class org.json4s.scalap.scalasig.MethodType)
	at org.json4s.reflect.ScalaSigReader$.getType$1(ScalaSigReader.scala:159)
	at org.json4s.reflect.ScalaSigReader$.findArgTypeForField(ScalaSigReader.scala:167)
	at org.json4s.reflect.ScalaSigReader$.readField(ScalaSigReader.scala:59)
	at org.json4s.reflect.Reflector$ClassDescriptorBuilder.$anonfun$fields$3(Reflector.scala:77)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at org.json4s.reflect.Reflector$ClassDescriptorBuilder.fields(Reflector.scala:76)
	at org.json4s.reflect.Reflector$ClassDescriptorBuilder.fields(Reflector.scala:89)
	at org.json4s.reflect.Reflector$ClassDescriptorBuilder.properties(Reflector.scala:93)
	at org.json4s.reflect.Reflector$ClassDescriptorBuilder.result(Reflector.scala:221)
	at org.json4s.reflect.Reflector$.createDescriptorWithFormats(Reflector.scala:61)
[...]

PR #291, opened 2021-08-26
Bugfix/agent 286 dsv2 databricks issues

### LINKED ISSUE #286: Support for `AppendData` node
ref https://github.com/AbsaOSS/spline-spark-agent/issues/256#issuecomment-901558208

cerveada:
fixes #286, #290

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
@@ -22,8 +22,9 @@ import za.co.absa.commons.lang.OptionImplicits._
 import za.co.absa.spline.harvester.ComponentCreatorFactory
 import za.co.absa.spline.harvester.ModelConstants.OperationExtras
 import za.co.absa.spline.harvester.builder.OperationNodeBuilder
+import za.co.absa.spline.harvester.converter.IOParamsConverter
 import za.co.absa.spline.harvester.postprocessing.PostProcessor
-import za.co.absa.spline.producer.model.v1_1.WriteOperation
+import za.co.absa.spline.producer.model.v1_1.{Attribute, WriteOperation}
 
 class WriteNodeBuilder
 (command: WriteCommand)
@@ -33,6 +34,8 @@ class WriteNodeBuilder
   override protected type R = WriteOperation
   override val operation: LogicalPlan = command.query
 
+  protected lazy val ioParamsConverter = new IOParamsConverter(exprToRefConverter)
+
   override def build(): WriteOperation = {
     val Seq(uri) = command.sourceIdentifier.uris
     val wop = WriteOperation(
@@ -41,7 +44,7 @@ class WriteNodeBuilder
       id = operationId,
       name = command.name.asOption,
       childIds = childIds,
-      params = Map(command.params.toSeq: _*).asOption,
+      params = ioParamsConverter.convert(command.params).asOption, // TODO why was not used the original map in the previous code?
       extra = Map(
         OperationExtras.DestinationType -> command.sourceIdentifier.format
       ).asOption
@@ -49,4 +52,6 @@ class WriteNodeBuilder
 
     postProcessor.process(wop)
   }
+
+  def additionalAttributes: Seq[Attribute] = attributeConverter.values
 }
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
@@ -44,7 +44,7 @@ class WriteNodeBuilder
       id = operationId,
       name = command.name.asOption,
       childIds = childIds,
-      params = ioParamsConverter.convert(command.params).asOption, // TODO why was not used the original map in the previous code?
+      params = ioParamsConverter.convert(command.params).asOption,
       extra = Map(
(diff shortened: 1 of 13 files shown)


integration-tests/src/main/scala/za/co/absa/spline/test/fixture/spline/SplineFixture.scala
```diff
@@ -34,6 +34,8 @@ trait SplineFixture {
 object SplineFixture {
   def EmptyConf = new BaseConfiguration
 
-  def extractTableIdentifier(paramsOption: Option[Map[String, Any]]): TableIdentifier =
-    paramsOption.get("table").asInstanceOf[Map[String, _]]("identifier").asInstanceOf[TableIdentifier]
+  def extractTableIdentifier(paramsOption: Option[Map[String, Any]]): Map[String, Any] =
```
wajda, line 37: Why the `TableIdentifier` isn't used anymore?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/AttributeReorderingFilter.scala
```diff
@@ -33,7 +33,7 @@ class AttributeReorderingFilter(conf: Configuration) extends AbstractPostProcess
   override def processExecutionPlan(plan: ExecutionPlan, ctx: HarvestingContext): ExecutionPlan = {
     val isByName = plan
       .operations.write.params
-      .flatMap(_.get("isByName").map(_.asInstanceOf[Boolean]))
+      .flatMap(_.get("isByName").map(_.asInstanceOf[Option[Boolean]].get))
```
wajda: Where did the `Option` come from. and why explicit `.get`?
Did you mean `.asInstanceOf[Option[Boolean]]` instead of `.map(_.asInstanceOf[Option[Boolean]])` ?

wajda: Since `"isByName"` literal is used in more than one place while referring the same thing it should be defined as a constant.

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DataSourceV2Plugin.scala
```diff
@@ -70,39 +71,19 @@ class DataSourceV2Plugin(
...
-    case `_: V2CreateTablePlan`(ctp) => {
-      val catalog = extractFieldValue[AnyRef](ctp, "catalog")
-      val identifier = extractFieldValue[AnyRef](ctp, "tableName")
-      val loadTableMethods = catalog.getClass.getMethods.filter(_.getName == "loadTable")
-      val table = loadTableMethods.flatMap(m => Try(m.invoke(catalog, identifier)).toOption).head
-      val sourceId = extractSourceIdFromTable(table)
-
-      val query = extractFieldValue[LogicalPlan](ctp, "query")
+    case `_: CreateTableAsSelect`(ctc) =>
+      val prop = "ignoreIfExists" -> extractFieldValue[Boolean](ctc, "ignoreIfExists")
```
wajda, line 75: I think we need to start respecting constants more.

core/src/main/scala/za/co/absa/spline/harvester/converter/IOParamsConverter.scala
```diff
@@ -0,0 +1,39 @@
...
+  override type To = Map[String, _]
+
+  private def valueDecomposer = ValueDecomposer.addHandler(_ => {
+    case (exp: Expression, _) => Some(exprToRefConverter.convert(exp))
+    case (ti: TableIdentifier, _) => Some(Map("table" -> ti.table, "database" -> ti.database))
+  })
+
+  override def convert(operation: Map[String, _]): Map[String, _] = {
+    operation.map{ case (k, v) =>
+      k -> valueDecomposer.decompose(v, "dummyType")
```
wajda: what's a dummy type?

core/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
```diff
@@ -41,12 +44,14 @@ class WriteNodeBuilder
       id = operationId,
       name = command.name.asOption,
       childIds = childIds,
-      params = Map(command.params.toSeq: _*).asOption,
+      params = ioParamsConverter.convert(command.params).asOption, // TODO why was not used the original map in the previous code?
```
wajda: That's fine, we don't need a `Map` wrapper here.

It's a leftover from the earlier versions where there was no `extra`, and `params` were used for carrying Spline specific things as well, hence extra keys were added to the map via `+` or `++`.
The problem used to be that since Spark maps are often case-insensitive, the resultant map could also be case-insensitive. That fact was very hard to notice and created an unwanted effect on the serialized DTOs where all added keys could appear unexpectedly lowercased.

Since we don't explicitly add any keys to `params` anymore, we can safely get rid of the wrapper.

integration-tests/src/main/scala/za/co/absa/spline/test/fixture/spline/SplineFixture.scala
```diff
@@ -34,6 +34,8 @@ trait SplineFixture {
...
+    paramsOption
+      .get("table").asInstanceOf[Some[Map[String, _]]]
+      .get("identifier").asInstanceOf[Map[String, _]]
```
wajda: Why `Some[Map[...]]`?

wajda (approved):

Issue #277, opened 2021-08-05
Google BigQuery Plugin


mohanajuhi166:
We have a requirement to add google Bigquery support  as part of the Plugin API support: https://github.com/AbsaOSS/spline-spark-agent#plugins.

Is there a possibility for this to be added as a feature?
I have did some development on my own as well regarding Bigquery.

Thanks & Regards.

wajda:
Sure, why not.
I would recommend you to create a plugin first, make it working on your side, and then create a PR when it's ready. Please also write some integration tests. 
Thank you.

mohanajuhi166:
Thank you @wajda ,
was working on the plugin and noticed that JDBCPlugin may work in this case for bigquery.
An example below:-
`spark.sqlContext.read.format("jdbc").option("url", "jdbc:googlebigquery:DataSetId=MyDataSetId;ProjectId=MyProjectId;").option("dbtable","Orders").option("driver","cdata.jdbc.googlebigquery.GoogleBigQueryDriver").load()`

Any thoughts on the same? Thank you.

wajda:
> Any thoughts on the same? Thank you.

I have nothing to add. If it works it works :)

mohanajuhi166:
PR raised : https://github.com/AbsaOSS/spline-spark-agent/pull/292

mohanajuhi166:
Raised a PR @wajda . I am not entirely sure what kind of integration tests i can write for bigquery.

wajda:
Can we create a test account and use Google free tier for automated testing purposes?

PR #294, opened 2021-08-30
Feature Request #277 : Bigquery plugin (Plugins API) 

mohanajuhi166:
Feature Link : #277

We have a requirement to add google BigQuery support as part of the Plugin API support: https://github.com/AbsaOSS/spline-spark-agent#plugins.

Attached #277
1). BigQuery Plugin code - captures - project, schema and table name
2). Required Dependencies.

Tested locally. Limitation to attach screenshot due to organizational reasons.

Can you please suggest me,what kind of integration tests, will we require in this regard?

Thanks &Regards,
Mohana

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -101,6 +101,18 @@
             <version>1.12.5</version>
         </dependency>
         <dependency>
+            <groupId>com.google.cloud.spark</groupId>
+            <artifactId>spark-bigquery_${scala.binary.version}</artifactId>
+            <version>0.22.0</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud.spark</groupId>
+            <artifactId>spark-bigquery-with-dependencies_${scala.binary.version}</artifactId>
+            <version>0.22.0</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
             <groupId>com.datastax.spark</groupId>
             <artifactId>spark-cassandra-connector_${scala.binary.version}</artifactId>
             <optional>true</optional>
--- a/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
```diff
@@ -0,0 +1,68 @@
...
+      val schema = cmd.options("schema")
+      val table = cmd.options("table")
+      val project = cmd.options("project")
+      (asSourceId(project, schema, table), cmd.mode, cmd.query, cmd.options)
+  }
+}
+
+object BigQueryPlugin {
+
+  object `_: BigQuerySourceRelation` extends SafeTypeMatchingExtractor[AnyRef]("com.google.cloud.spark.bigquery.direct.DirectBigQueryRelation")
```
wajda, line 62:
```suggestion
  object `_: DirectBigQueryRelation` extends SafeTypeMatchingExtractor[AnyRef]("com.google.cloud.spark.bigquery.direct.DirectBigQueryRelation")
```
I understand that it might be hardly evident, but we use that back tick naming trick to mimic conventional Scala type based pattern matching, that we would normally use if the matching class were guaranteed to be available at runtime. So, the identifier after `_:` must be the exact class name that is being matched - `DirectBigQueryRelation`

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2020 ABSA Group Limited
```
wajda, line 2:
```suggestion
 * Copyright 2021 ABSA Group Limited
```

core/pom.xml
```diff
@@ -100,6 +100,18 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud.spark</groupId>
+            <artifactId>spark-bigquery_${scala.binary.version}</artifactId>
+            <version>0.22.0</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud.spark</groupId>
+            <artifactId>spark-bigquery-with-dependencies_${scala.binary.version}</artifactId>
```
wajda, line 111: This dependency is not used.

wajda (commented):
Thank you for the PR!
It looks good to me excepts for a few minor things mentioned.

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
cerveada: Could you keep the line's length under 120?

cerveada:
Integration test would be nice, but I don't know if there are some tools to mock a Bigquery?

If you want some test examples, take a look on ExcelSpec. There should be only one lineage produced inside each `captor.lineageOf` and the test asserts should happen in the `yield` block. We have to use this because the spark lineage events produced are asynchronous.

wajda:
closing it in favor of #313 which is a backport of this change to the _release/0.6_ branch + minor fixes.

PR #301, opened 2021-09-03
spark agent #289 improve write extraction fail logs

### LINKED ISSUE #289: Redesing the behaviour when Write command is not recognized
- Currently, the message that appears in logs differs depending on spline mode, which is confusing.
- When there is an error, it could be a good idea to try to use structure dumper (but catch any errors form there)
- generaly make the logs the same in the same sittuation (only additional details hould be added / missing)

This happens in #286

cerveada:
Fixes #289

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
@@ -35,11 +35,11 @@ trait ObjectStructureLogging {
 
   private def logObjectStructure(
     obj: AnyRef,
-    logMethod: String => Unit,
-    logMethodThrowable: (String, Throwable) => Unit
+    logFunction: String => Unit,
+    logFunctionThrowable: (String, Throwable) => Unit
   ): Unit =
     Try(ObjectStructureDumper.dump(obj)) match {
-      case Success(s) => logMethod(s)
-      case Failure(e) if NonFatal(e) => logMethodThrowable(s"Attempt to dump structure of ${obj.getClass} failed", e)
+      case Success(s) => logFunction(s)
+      case Failure(e) => logFunctionThrowable(s"Attempt to dump structure of ${obj.getClass} failed", e)
     }
 }
--- a/core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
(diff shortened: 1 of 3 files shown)


core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
```diff
@@ -0,0 +1,45 @@
...
+    if (log.isErrorEnabled) logObjectStructure(obj, log.error, log.error)
+
+  private def logObjectStructure(
+    obj: AnyRef,
+    logMethod: String => Unit,
+    logMethodThrowable: (String, Throwable) => Unit
+  ): Unit =
+    Try(ObjectStructureDumper.dump(obj)) match {
+      case Success(s) => logMethod(s)
+      case Failure(e) if NonFatal(e) => logMethodThrowable(s"Attempt to dump structure of ${obj.getClass} failed", e)
```
wajda: When using `Try` checking for `NonFatal` is redundant. `Try` only catches non-fatal errors by definition.

core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.spark.internal.Logging
+
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
+
+trait ObjectStructureLogging {
+  self: Logging =>
+
+  protected def logTraceObjectStructure(obj: => AnyRef): Unit =
+    if (log.isTraceEnabled) logObjectStructure(obj, log.trace, log.trace)
```
wajda, line 28: Though `log` has a type of `slf4j.Logger` the trait itself is combined with `spark.Logging` that has `logXXX()` methods doing the same thing. So we can use them directly instead of the `log` (that is one level more distant dependency).
```suggestion
    logObjectStructure(obj, logTrace, logTrace)
```

core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureLogging.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  protected def logWarningObjectStructure(obj: => AnyRef): Unit =
+    if (log.isWarnEnabled) logObjectStructure(obj, log.warn, log.warn)
+
+  protected def logErrorObjectStructure(obj: => AnyRef): Unit =
+    if (log.isErrorEnabled) logObjectStructure(obj, log.error, log.error)
+
+  private def logObjectStructure(
+    obj: AnyRef,
+    logMethod: String => Unit,
```
wajda: naming: it might sound picky but sill - `logMethod` is not a method but a function. It confused me for a second when I read it like "log the method ...". I'd name it `logFn`

wajda (approved):

PR #325, opened 2021-10-13
Feature/agent 169 declarative extra metadata

### LINKED ISSUE #169: Declarative way for adding user extra metadata
Related to AbsaOSS/spline#737 and extending #35

The idea is to create a mechanism through which users could declaratively control what additional info should be captured (or setting constant values like tags etc) under the `extra` property of the execution plan and the execution event.
This should be achievable (in majority simple cases) without any code change, purely via the configuration.

Imaginary example:
```properties
spline.userExtraMetadata.injectRules = 
    { 
        "executionPlan": {
            "foo": { "$eval": "$session.config.foo" },
            "bar": { "$env": "BAR_HOME" },
            "baz": { "$jvm": "some.jvm.prop" },
            "qux": 42,
            "tags": ["aaa", "bbb", "ccc"]
        },
        "operation[nodeName=abc]": {
            "z": { "$eval": "$op.x.y.z" }
        },
        "executionEvent": {
            "...": ...
        }
    }
```

Applying the above config would result in `execPlan.extra` to have properties `foo`, `bar`, `baz` with values obtained respectively from evaluating an expression `$session.config.foo`, a value of `BAR_HOME` environment variable and a value of a `"some.jvm.prop"` JVM property (set via `-D`). The extra properties `qux` and `tags` would be just given constant values.
All operations whose `nodeName` equals to `"abc"` would have an extra `z` property holding a result of evaluating the given expression.

cerveada:
Extra template example:
```json
{
    "executionPlan": {
        "qux": 42,
        "tags": ["aaa", "bbb", "ccc"],
        "foo": { "$js": "executionPlan.name()" },
        "bar": { "$env": "BAR_HOME" },
        "baz": { "$jvm": "some.jvm.prop" },
        "daz": { "$js": "session.conf().get('k')" }
    }
}
```
Extras from all filters are merged deeply. For example the tags sequences will be unionized not overwritten.

Example of optional predicate:
```json
{
    "executionEvent[@.extra['foo'] == 'a' && @.extra['bar'] == 'x']": {
        "bux": 2
    }
}
```
The current entity (executionEvent in the example) is available as `@` other available objects are: `logicalPlan`, `executedPlanOpt` and `session`.

When traversing the path, the optional values are automatically unwrapped and maps are accessed via keys. 
There is alternative map accesor `map['the.key']` that can be used to access arrays as well: `seq[2]`.

See the tests for more examples.

---

Example of jvm props config
```
-Dspline.lineageDispatcher=logging
-Dsome.jvm.prop=42
-Dspline.userExtraMetadata.injectRules"={\"executionPlan\":{\"foo\":{\"$js\":\"executionPlan.name()\"}\,\"bar\":{\"$env\":\"JAVA_HOME\"}\,\"baz\":{\"$jvm\":\"some.jvm.prop\"}\,\"qux\":42\,\"tags\":[\"aaa\"\,\"bbb\"\,\"ccc\"]}}"
```
`"` and `,` must be escaped. This is quite cumbersome, so I added:

```
-Dspline.userExtraMetadata.injectRules.file=/path/to/file.json
```


core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -121,8 +123,18 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
...
+    Seq(new AttributeReorderingFilter(), new OneRowRelationFilter()) ++ createDeclarativeExtraInjectingFilterOption()
+  }
+
+  private def createDeclarativeExtraInjectingFilterOption(): Option[PostProcessingFilter] =
+    Try(DeclarativeExtraInjectingFilter(configuration)) match {
+      case Success(maybeFilter) => maybeFilter
+      case Failure(e) if splineMode == SplineMode.REQUIRED => throw e
+      case Failure(e) if splineMode == SplineMode.BEST_EFFORT =>
+        logWarning("DeclarativeExtraInjectingFilter initialization failed.", e)
+        None
```
wajda: This error handling is already in place upper on the stack in the method `QueryExecutionEventHandlerFactory.withErrorHandling()`. Here there there is no need to catch anything.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package za.co.absa.spline.harvester.postprocessing.extra
+
+import org.apache.spark.internal.Logging
+
+import javax.script.ScriptEngine
+import scala.util.{Failure, Success, Try}
+
+object ExtraTemplateModel extends Logging{
```
wajda: nit: white space before `{`

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  class ExtraTemplate(extra: Map[String, Any]) {
+
+    def eval(bindings: Map[String, Any]): Map[String, Any] =
+      Try(tryEval(bindings)) match {
+        case Success(value) => value
+        case Failure(exception) =>
+          logWarning("DeclarativeExtraInjectingFilter's template evaluation failed, " +
+            "filter will not be applied.", exception)
```
wajda: code layout: unnecessary string concatenation, the message doesn't seem to be too long.
Also, if all arguments don't fit one line, they should be arrange one argument per line.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import javax.script.ScriptEngine
+import scala.util.{Failure, Success, Try}
+
+object ExtraTemplateModel extends Logging{
+
+  class ExtraTemplate(extra: Map[String, Any]) {
+
+    def eval(bindings: Map[String, Any]): Map[String, Any] =
+      Try(tryEval(bindings)) match {
```
wajda: Suggestion - you don't call any monadic operation on the `Try`, so here it's created only to be pattern-matched immediately.
I would use normal `try-catch` in this case. The code would be shorter.
Another thought - method `tryEval` could return `Try` instead of throwing. Then the above pattern-match would make more sense.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+          logWarning("DeclarativeExtraInjectingFilter's template evaluation failed, " +
+            "filter will not be applied.", exception)
+          Map.empty
+      }
+
+    private def tryEval(bindings: Map[String, Any]): Map[String, Any] =
+      extra.transform((k, v) => evalValue(v, bindings))
+
+    private def evalValue(value: Any, bindings: Map[String, Any]): Any = value match {
+      case m: Map[String,_] => m.transform((k, v) => evalValue(v, bindings))
```
wajda: nit: missing while space after comma

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  case class EnvVar(envName: String) extends Evaluable {
+    override def eval(bindings: Map[String, Any]): AnyRef = System.getenv(envName)
+  }
+
+  case class JsEval(jsEngine: ScriptEngine, js: String) extends Evaluable {
+    override def eval(bindings: Map[String, Any]): AnyRef = {
+
+      val jsBindings = jsEngine.createBindings
+      bindings.foreach{ case (k, v) => jsBindings.put(k, v) }
```
wajda: nit: missing while space before `{`

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    private def evalValue(value: Any, bindings: Map[String, Any]): Any = value match {
+      case m: Map[String,_] => m.transform((k, v) => evalValue(v, bindings))
+      case s: Seq[_] => s.map(evalValue(_, bindings))
+      case e: Evaluable => e.eval(bindings)
+      case v => v
+    }
+  }
+
+  trait Evaluable {
```
wajda: shouldn't the trait to be sealed?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
wajda: The object seems to be only used as a namespace. Why don't simply put its content into the `model` sub-package and remove the object?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/DeclarativeExtraInjectingFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+
+class DeclarativeExtraInjectingFilter(
+  planExtraDefs: Seq[ExtraDef],
+  eventExtraDefs: Seq[ExtraDef],
+  readExtraDefs: Seq[ExtraDef],
+  writeExtraDefs: Seq[ExtraDef],
+  operationExtraDefs: Seq[ExtraDef]
+) extends PostProcessingFilter with Logging {
+
+  override def processExecutionEvent(event: ExecutionEvent, ctx: HarvestingContext): ExecutionEvent = {
```
wajda, line 42: curly braces: this method is wrapped with the curly braces, but the others aren't. It should be consistent. (Personally I would not omit curly braces of a several line method body with indentation. It helps seeing method boundaries better.)

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/DeclarativeExtraInjectingFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  private def mergeValues(v1: Any, v2: Any): Any = (v1, v2) match {
+    case (v1: Map[String, _], v2: Map[String, _]) => deepMergeMaps(v1, v2)
+    case (v1: Seq[Any], v2: Seq[Any]) => (v1 ++ v2).distinct
+    case (v1, v2) => v2
+  }
+
+}
+
+object DeclarativeExtraInjectingFilter extends Logging {
```
wajda, line 114: unused `Logging`

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/DeclarativeExtraInjectingFilter.scala
```diff
@@ -0,0 +1,153 @@
...
+    val maybeExtraDefMap = conf
+      .getOptionalString(InjectRulesKey)
+      .orElse(conf.getOptionalString(s"$InjectRulesKey.$FileKey").map(loadFromFile))
+      .map(_.fromJson[Map[String, Map[String, Any]]])
+
+    maybeExtraDefMap
+      .map(createFilter)
+  }
+
+  private def loadFromFile(path: String): String = {
```
wajda: `org.apache.commons.io.IOUtils` has a bunch of overloaded `toString(...)` methods that do the same thing as this method. Think about reusing it.
Also, why to use _file_ as a config source? Can it be URL instead? It would be much more flexible.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateParser.scala
```diff
@@ -0,0 +1,46 @@
...
+package za.co.absa.spline.harvester.postprocessing.extra
+
+
+import za.co.absa.spline.harvester.postprocessing.extra.ExtraTemplateModel._
+
+import javax.script.ScriptEngineManager
+
+object ExtraTemplateParser {
+
+  private val jsEngine = new ScriptEngineManager().getEngineByMimeType("text/javascript")
```
wajda: According to the `jsEngine.getFactory.getParameter("THREADING")` return value `null` the engie instance isn't thread-safe (see https://docs.oracle.com/javase/8/docs/api/javax/script/ScriptEngineFactory.html#getParameter-java.lang.String-)
Although in practice, the race-condition should not occur (due to the fact that Spline harvester is invoked from the Spark event thread in a synchronous manner), it nevertheless is not a good practice if a utility class like this makes such assumptions about the global environment and usage specifics.
The thread-safeness should be either guaranteed by the implementation, or otherwise communicated to and managed by the caller of the `parse()` method.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateParser.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private def parseRec(v: Any): Any = v match {
+    case m: Map[String,_] if m.contains(EvaluableNames.JVMProp) =>
+      JVMProp(m(EvaluableNames.JVMProp).asInstanceOf[String])
+
+    case m: Map[String,_] if m.contains(EvaluableNames.EnvVar) =>
+      EnvVar(m(EvaluableNames.EnvVar).asInstanceOf[String])
+
+    case m: Map[String,_] if m.contains(EvaluableNames.JsEval) =>
+      JsEval(jsEngine, m(EvaluableNames.JsEval).asInstanceOf[String])
```
wajda: Does it mean that if the map contains any of those three keys then only the one that matches first would be processed and the rest of map entries would be ignored?

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
cerveada: The Idea was to not stop agent when the user just provided bad config for the filter, but maybe it's better to fail the whole agent.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateParser.scala
cerveada: Yes, it should contain just one key when it is evaluable, I can add I check for that.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraTemplateModel.scala
cerveada: Yes it would be shorter, but the readability seems to me will not change. I think there is not much difference.

I tend to use Try because it handles the Fatal exceptions automatically.

wajda (approved):
LGTM

PR #327, opened 2021-10-14
Feature/323 UUID v5

wajda:
- Introduces `IdGenerator` trait that is used to generate ID for all Spline model components
- added support for UUID v3 and v5
- `ExecutionPlan` is set to UUID v5 by default, but can be configured `spline.internal.execPlan.uuid.version`
- `DataType` uses a composite ID generator - UUID5 on a number sequence
- rest components use sequential ID generators
- added prefixes to operation. attribute and expression IDs - `op-`, `attr-` and `expr-` respectively for the sake of readability (for easier debugging and troubleshooting of execution plans)
- Exclude `spark.Expression.exprId` property from capturing (it breaks Harvester determinism)

Other  changes:
- fixed `durationNs` property type in the Swagger definition
- remove redundant `lazy` from some vals in `DefaultSplineConfigurer`
- `HarvestingContext` is now a normal class
- Got rid of `ComponentCreatorFactory` due to redundancy


core/src/main/scala/za/co/absa/spline/harvester/IdGenerators.scala
```diff
@@ -0,0 +1,40 @@
...
+    case v => throw new IllegalArgumentException(s"UUID version $v is not supported")
+  }
+
+  val attributeIdGenerator: IdGenerator[Any, String] = new SequentialIdGenerator("attr-{0}")
+  val expressionIdGenerator: IdGenerator[Any, String] = new SequentialIdGenerator("expr-{0}")
+  val operationIdGenerator: IdGenerator[Any, String] = new SequentialIdGenerator("op-{0}")
+  val dataTypeIdGenerator: IdGenerator[Any, UUID] =
+    new ComposableIdGenerator(
+      new SequentialIdGenerator("{0}"),
+      new UUID5IdGenerator[String](UUIDNamespace.DataType)
```
cerveada, line 32: Why is the SequentialIdGenerator needed here?

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -42,6 +43,14 @@ object DefaultSplineConfigurer {
      */
     val Mode = "spline.mode"
 
+    /**
+     * The UUID version that is used for ExecutionPlan ID.
+     * Note: Hash based versions (3 and 5) produce deterministic IDs based on the ExecutionPlan body.
+     *
+     * @see [[SplineMode]]
```
cerveada: ??

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -79,15 +88,15 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
 
   import collection.JavaConverters._
 
-  private lazy val configuration = new CompositeConfiguration(Seq(
+  private val configuration = new CompositeConfiguration(Seq(
```
cerveada, line 90: Are you sure this will work for both in code and codeless initialization without being lazy?

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
```diff
@@ -96,6 +105,12 @@ class DefaultSplineConfigurer(sparkSession: SparkSession, userConfiguration: Con
     }
   }
 
+  private val execPlanUUIDVersion: UUIDVersion = {
+    val uuidVer = configuration.getRequiredInt(ExecPlanUUIDVersion)
+    new IdGenerators(uuidVer) // validation
```
cerveada: It seems strange to create the generators just for validation.
`UUIDVersion` could be an actual class and the validation could happen in the constructor

core/src/main/scala/za/co/absa/spline/harvester/IdGenerators.scala
wajda: Because we want to generate unique (per generator instance) IDs, but we don't want them to be random.
The sequence number serves as an input for the hash based UUID generator.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
wajda: A copy-paste artifact. Thanks.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
wajda: According to our integration tests as well as example, yes it works. But I'll do more testing to be completely sure.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
wajda: I wanted to have a single place where an integer version number is mapped to a UUID generator implementation and this seemed as the easiest approach. The generators are created per harvester instance, but here we are at the agent config stage. So basically, here I need to check if the number is in the range, and later I need to instantiate a generator based on that number. To not duplicate the information about available UUID versions in two places, I tried reflective approach (discovering the implementation and instantiating it dynamically), factory approach (similar to what you suggest - create a proper factory on the config stage and use it later to instantiate a generator), but all approaches seemed as overkill for such a simple thing. So I decided that simply instantiating the generators facade one more time at the config time and throwing it away is not a big deal and serves the purpose of validation (like a dry-run), and it's a single line of code.

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
wajda: removed

core/src/main/scala/za/co/absa/spline/harvester/conf/DefaultSplineConfigurer.scala
wajda: Refactored with the factory approach, but made it a `Function1` type alias instead of a trait

cerveada (approved):

PR #338, opened 2021-10-22
update examples/README

wajda:
- fix run examples as a docker readme
- add build docker info


README.md
```diff
@@ -465,6 +465,19 @@ Spline will pick it up automatically.
 <a id="building"></a>
 ### Building for different Scala and Spark versions
 
+**Note:** The project requires Java version 1.8 (strictly) and [Apache Maven](https://maven.apache.org/) for building.
```
cerveada, line 468: Why Java 11 does not work?

cerveada (approved):

README.md
wajda: Because of some test dependencies, I don't remember exactly at the moment, but it was discussed X times before. Agent works on any Java supported by Spark, but requires strictly 1.8 for building.

PR #340, opened 2021-10-25
spark agent #335 remove deprecated UserExtraMetadataProvider

### LINKED ISSUE #335: Remove deprecated UserExtraMetadataProvider
- UserExtraMetadataProvider is depricated some time already
- We have DeclarativeExtraInjectingFilter as a replacement

cerveada:


integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
```diff
@@ -332,9 +330,31 @@ class LineageHarvesterSpec extends AsyncFlatSpec
       }
     }
 
+
+  private val injectRules =
```
wajda: this variable is defined on the class level, but is only used in one test. Consider either moving it closer to the usage, or (if the semantics is generic enough to be potentially useful for other test scenarios) renaming it to better reflect the purpose and moving it into a companion object.

cerveada: Moved to the test method

wajda (approved):

PR #355, opened 2021-11-12
issue #32 Capture failed executions

### LINKED ISSUE #32: Capture failed executions
If a job fails we should still send an execution plan and the error to the server.
Failed events should be filtered out in the query when building a lineage across jobs.

wajda:
- Capture failed execution events
- Add `spline.sql.failure.capture` config enum property with values `NONE`, `NON_FATAL` and `ALL`, that controls what kind of failed executions should be captured (respectively _none_, only `NonFatal` and _any_). Default mode is `NON_FATAL`.
- Add a new _Producer API_ protocol version number _1.2_ that is technically identical to _1.1_ and is only used to distinguish between the server implementation that supports and doesn't support failed execution capturing. The reason for that is that if a new agent is used against the server that doesn't properly support failed execution capturing (0.7.5 or earlier) even though it is defined in a protocol, then the failed events would be processed as if they were successful. That would subsequently result in displaying a wrong end-to-end lineage on the UI. To not force users to upgrade their server the newer agent should mimic an older behavior and not dispatch failed executions over _Producer API_ protocol ver _1.1_ or earlier.
- Add a failing example job

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
@@ -20,13 +20,12 @@ import za.co.absa.spline.harvester.dispatcher.LineageDispatcher
 
 import scala.concurrent.duration._
 import scala.language.postfixOps
-import scala.util.Try
 
 class QueryExecutionEventHandler(
   harvesterFactory: LineageHarvesterFactory,
   lineageDispatcher: LineageDispatcher) {
 
-  def handle(qe: QueryExecution, result: Try[Duration]): Unit = {
+  def handle(qe: QueryExecution, result: Either[Throwable, Duration]): Unit = {
     harvesterFactory
       .harvester(qe.analyzed, Some(qe.executedPlan))
       .harvest(result)
--- a/core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
@@ -15,11 +15,8 @@
 
 package za.co.absa.spline.harvester
 
-import java.util.UUID
-
 import org.apache.spark.sql.execution.QueryExecution
 import za.co.absa.spline.harvester.dispatcher.LineageDispatcher
-import za.co.absa.spline.harvester.json.HarvesterJsonSerDe.impl._
 
 import scala.concurrent.duration._
 import scala.language.postfixOps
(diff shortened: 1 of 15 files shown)


core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
```diff
@@ -29,32 +26,14 @@ class QueryExecutionEventHandler(
...
 
-  /**
-   * The method is executed when an action execution is successful.
-   *
-   * @param funcName   A name of the executed action.
-   * @param qe         A Spark object holding lineage information (logical, optimized, physical plan)
-   * @param durationNs Duration of the action execution [nanoseconds]
-   */
-  def onSuccess(funcName: String, qe: QueryExecution, durationNs: Long): Unit = {
+  def handle(qe: QueryExecution, result: Try[Duration]): Unit = {
```
cerveada: - It seems to me the main result here is the QueryExecution not a duration (maybe the result should be renamed to duration)
- The Try seems to suggest the error happened when generating the duration, not durring the whole job execution

Maybe in this case `Either` would be better or two params instead of one Try?

core/src/main/scala/za/co/absa/spline/harvester/conf/SplineConfigurer.scala
```diff
@@ -40,6 +41,12 @@ trait SplineConfigurer {
     * @return [[SplineMode]]
     */
   def splineMode: SplineMode
+
+  /**
+   * Controls if failed SQL execution should be captured depending on the kind of associated error
+   * @return
```
cerveada:
```suggestion
   * @return [[SQLFailureCaptureMode]]
```

core/src/main/scala/za/co/absa/spline/harvester/QueryExecutionEventHandler.scala
wajda: done

cerveada (approved):

diegoeche:
This is so awesome, we were waiting for this! :D

wajda:
soon, soon... :) This is in a develop branch, there are a few more things scheduled for that version and I'm on full-time training for the next two weeks, so no time for coding. Hopefully around Xmas time we'll release something

Note that this PR is for capturing errors that happen during job executions when the exec plan is known. Depending on how the agent is implemented, in Spark for instance a lot of logical errors are caught by the analyzer and are thrown before the exec plan is created. Those errors aren't captured by this feature.

PR #360, opened 2021-11-24
issue #359 Simplify programmatic initialization API

### LINKED ISSUE #359: Improve programmatic initialization API
The programmatic init API should provide a concise and elegant way to pass config properties, filters, dispatchers and other components to the Spline context.

```scala
spark.enableLineageTracking(
  AgentConfig.builder()
    .splineMode(SplineMode.DISABLED)
    .sqlFailureCaptureMode(SQLFailureCaptureMode.ALL)
    .postProcessingFilter(new PostProcessingFilter { /* my custom filter */ })
    .lineageDispatcher(new LineageDispatcher { /* my custom dispatcher */})
    .config("foo", 42)
    .config("bar", 777)
    .build()
)
```

wajda:
Enable the following API (tried to make it looking similar to Spark session builder API, familiar to developers) :
```scala
spark.enableLineageTracking(
  AgentConfig.builder()
    .splineMode(SplineMode.DISABLED)
    .sqlFailureCaptureMode(SQLFailureCaptureMode.ALL)
    .postProcessingFilter(new PostProcessingFilter { /* my custom filter */ })
    .lineageDispatcher(new LineageDispatcher { /* my custom dispatcher */})
    .config("foo", 42)
    .config("bar", 777)
    .build()
)
```

Besides the changes and refactoring needed for API changing I have made the following fixes and improvements along the way:
  - Fix `sqlFailureCapturingMode` to apply in codeless init as well.
  - Renamed `DeclarativeExtraInjectingFilter` to `ExtraMetadataCapturingFilter` and made it a normal filter configured via properties.
  - Added a new `name` property to all post-processing filters, dispatchers and strategies for better logging.
  - Got rid of `NoOpLineageDispatcher` and simplified `SplineFixture`
  - Consolidated all initialization and listener registering code in a single place (`SparkLineageInitializer`)
  - For some enums that are exposed in the user API, replaced Scala ones with Java ones to improve interoperability with Java clients.


README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
 The file [spline.default.properties](core/src/main/resources/spline.default.properties) contains default values 
 for all Spline properties along with additional documentation.
 It's a good idea to look in the file to see what properties are available.
 
+The order of precedence might look counter-intuitive, as one woudl expect that explicitly provided config (`AgentConfig` instance) should 
+override ones defined in the outer scope. However, prioritizing global config to local one makes it easier to manage Spline settings centrally
+on clusters, while still allowing room for customization by job developers.
+
+For example, a company could require lineage metadata from jobs executed on a particular cluster to be sanitized, enhanced with some metrics 
+and credentials and stored in a certain metadata store (a database, file, Spline server etc). The Spline configured needs to be set globally
```
cerveada:
```suggestion
and credentials and stored in a certain metadata store (a database, file, Spline server etc). The Spline configuration needs to be set globally
```

README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
+and applied to all Spark jobs automatically. However, some jobs might contain hardcoded properties that the developers used locally or on 
+a testing environment, and forgot to remove them before submitting jobs into a production.
+In such situation we want cluster settings to have precedence over the hardcoded ones.
```
cerveada: Could `hardcoded` be replaced by `local`?

README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
+Assuming that hardcoded settings would most likely be defined in thone of the `AgentConfig` object, a property file or a JVM properties, 
+on the cluster we could define them in the Spark config or Hadoop config.
+
+When a property is defined in multiple places, from the perspective of how the conflict is resolved, we can describe two kinds of Spline 
+configuration properties. For majority properties the "first occurrence wins" rule is applied. But there are two that are composed instead
+of been overridden: `spline.lineageDispatcher` and `spline.postProcessingFilter`. E.g. if a _LineageDispatcher_ is set to be _Kafka_ in one config
```
cerveada:
```suggestion
In case of multiple definitions of property the first occurrence wins, but `spline.lineageDispatcher` and `spline.postProcessingFilter` properties are composed instead. E.g. if a _LineageDispatcher_ is set to be _Kafka_ in one config
```

This sentence seems somewhat complicated to follow. Maybe something like this would work better?

README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
+The order of precedence might look counter-intuitive, as one woudl expect that explicitly provided config (`AgentConfig` instance) should 
+override ones defined in the outer scope. However, prioritizing global config to local one makes it easier to manage Spline settings centrally
+on clusters, while still allowing room for customization by job developers.
+
+For example, a company could require lineage metadata from jobs executed on a particular cluster to be sanitized, enhanced with some metrics 
+and credentials and stored in a certain metadata store (a database, file, Spline server etc). The Spline configured needs to be set globally
+and applied to all Spark jobs automatically. However, some jobs might contain hardcoded properties that the developers used locally or on 
+a testing environment, and forgot to remove them before submitting jobs into a production.
+In such situation we want cluster settings to have precedence over the hardcoded ones.
+Assuming that hardcoded settings would most likely be defined in thone of the `AgentConfig` object, a property file or a JVM properties,
```
cerveada: thone ?

core/src/main/scala/za/co/absa/commons/ConfigurationImplicits.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package za.co.absa.commons
+
+import org.apache.commons.configuration.Configuration
+import za.co.absa.commons.config.ConfigurationImplicits.ConfigurationRequiredWrapper
+
+object ConfigurationImplicits {
+
+  implicit class ConfOps(val conf: Configuration) extends AnyVal {
+    def getRequiredEnum[A <: Enum[A]](key: String, valueOfFn: String => A, values: Seq[A]): A = {
```
cerveada: Why `valueOfFn` and `values` needs to be provided. Aren't they methods that every enum has? Could the class be used?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/model/predicate/extraPredicateModel.scala
```diff
@@ -162,7 +162,9 @@ object ComparisonSymbol {
   val GTE = ">="
 }
 
-object BaseNodeNames {
+object BaseNodeName {
+  type Value = String
```
cerveada: What is the purpose of this?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraMetadataCapturingFilter.scala
```diff
@@ -19,74 +19,103 @@ package za.co.absa.spline.harvester.postprocessing.extra
...
+import scala.util.Try
 
-class DeclarativeExtraInjectingFilter(
-  planExtraDefs: Seq[ExtraDef],
-  eventExtraDefs: Seq[ExtraDef],
-  readExtraDefs: Seq[ExtraDef],
-  writeExtraDefs: Seq[ExtraDef],
-  operationExtraDefs: Seq[ExtraDef]
-) extends PostProcessingFilter {
+class ExtraMetadataCapturingFilter(allDefs: Map[BaseNodeName.Value, Seq[ExtraDef]]) extends PostProcessingFilter {
```
cerveada: `ExtraMetadataCapturingFilter` - It may not capture anything, though. It may just inject some static extra data, or it captures and injects a dynamic extra data.

examples/src/main/java/za/co/absa/spline/example/batch/JavaExampleJob.java
```diff
@@ -18,22 +18,33 @@
...
-            .write()
-            .mode(SaveMode.Overwrite)
-            .csv("data/output/batch/java-sample.csv");
+                .option("header", "true")
+                .option("inferSchema", "true")
+                .csv("data/input/batch/wikidata.csv")
+                .as("source")
+                .write()
+                .mode(SaveMode.Overwrite)
+                .csv("data/output/batch/java-sample.csv");
```
cerveada: Indentation. We use 2 spaces everywhere.

README.md
wajda: I thought about it, but by _local_ I would understand something that is local to a particular host. Is it correctly to say that, for instance, `System.setProperty(...)` is local settings?

core/src/main/scala/za/co/absa/commons/ConfigurationImplicits.scala
wajda: Concrete enum classes are generated by the java compiler. `values()` method is autogenerated and the `valueOf` is a static member on `java.lang.Enum`. So, on my awareness, there is no supertype of, e.g. `SplineEnum` that I can call those two methods on.
I could try reflective approach though.

README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
 - JVM system properties
 - `spline.properties` file on classpath
+- `AgentConfig` object
+- `spline.default.properties` file on classpath
 
 The file [spline.default.properties](core/src/main/resources/spline.default.properties) contains default values 
 for all Spline properties along with additional documentation.
 It's a good idea to look in the file to see what properties are available.
 
+The order of precedence might look counter-intuitive, as one woudl expect that explicitly provided config (`AgentConfig` instance) should
```
cerveada:
```suggestion
The order of precedence might look counter-intuitive, as one would expect that explicitly provided config (`AgentConfig` instance) should 
```

README.md
cerveada: I don't know. What about this:

```suggestion
In such situation we want cluster settings to have precedence over the job settings.
```

core/src/main/scala/za/co/absa/commons/ConfigurationImplicits.scala
cerveada: ok, then don't worry about it

wajda: done

README.md
wajda: a typo - removed

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/model/predicate/extraPredicateModel.scala
wajda: Singular vs plural. It resembles enum and enums are usually named with singular nouns. I wonder if it shouldn't be a proper enum?
Type alias helps to read method signatures when those constants are passed as arguments.

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraMetadataCapturingFilter.scala
wajda: I didn't get what you mean by this one. This post-processing filter captures the user extra metadata from different sources  as defined by custom rules. It's capturing, not injection.

examples/src/main/java/za/co/absa/spline/example/batch/JavaExampleJob.java
wajda: fixed

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/model/predicate/extraPredicateModel.scala
cerveada: It's a strange that `Value` is not value, it's a type of the names. Could it be `Type` instead?

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraMetadataCapturingFilter.scala
cerveada: If I want to just add some tags to the lineage, I can specify a rule to do that. In that case I do not capture anything, I am just **injecting** the extra tags.

In case I want to capture additional information from lineage, I can specify the rule to **captrue** it and than **injecting** as an extra data to lineage.

My point is the capturing happens only in some cases, but the extra data are always injected into the standard lineage, that is the purpose of the filter.

README.md
```diff
@@ -153,24 +153,66 @@ import za.co.absa.spline.harvester.SparkLineageInitializer;
...
+For example, a company could require lineage metadata from jobs executed on a particular cluster to be sanitized, enhanced with some metrics 
+and credentials and stored in a certain metadata store (a database, file, Spline server etc). The Spline configuration needs to be set globally
+and applied to all Spark jobs automatically. However, some jobs might contain hardcoded properties that the developers used locally or on 
+a testing environment, and forgot to remove them before submitting jobs into a production.
+In such situation we want cluster settings to have precedence over the hardcoded ones.
+Assuming that hardcoded settings would most likely be defined in the `AgentConfig` object, a property file or a JVM properties, 
+on the cluster we could define them in the Spark config or Hadoop config.
+
+When a property is defined in multiple places, from the perspective of how the conflict is resolved, we can describe two kinds of Spline 
+configuration properties. For majority properties the "first occurrence wins" rule is applied. But there are two that are composed instead
```
cerveada: I still think this is sentence could be reworded to make it clearer.
```suggestion
For a properties defined in multiple places Spline has two ways how to resolve the conflict: For majority of properties the "first occurrence wins" rule is applied. But there are two that are composed instead
```

cerveada: also
`instead of been overridden` -> `instead of being overridden`

core/src/main/scala/za/co/absa/spline/harvester/postprocessing/extra/ExtraMetadataCapturingFilter.scala
wajda: So, if I get it right we are merely talking about wording here. From what you describe I suggest the the word  '__collecting__'. The problem with the word '__injecting__' is that it puts too much emphasis on what it does with the extra metadata (injects it somewhere) rather than focusing on the primary goal which is to collect (capture) some extra stuff in addition to what is collected by default.

cerveada (approved):

Issue #370, opened 2021-12-22
Spark Config :: Arrays properties don't work


wajda:
I'm having trouble getting the composite filter working with two user-defined filters. I'm doing the Spline configuration in my PySpark application, so I have code that now looks like this:
```
        spark = SparkSession \
            .builder \
            .appName(app_name) \
            .config('spark.sql.queryExecutionListeners', 'za.co.absa.spline.harvester.listener.SplineQueryExecutionListener') \
            .config('spark.spline.lineageDispatcher', 'http') \
            .config('spark.spline.lineageDispatcher.http.producer.url', spline_config.url) \
            .config('spark.spline.mode', spline_config.mode.value) \
            .config('spark.spline.postProcessingFilter', 'composite') \
            .config('spark.spline.postProcessingFilter.composite.className', 'za.co.absa.spline.harvester.postprocessing.CompositePostProcessingFilter') \
            .config('spark.spline.postProcessingFilter.composite.filters', 'mcd,password') \
            ... (truncated for brevity)
```
where `password` is the `DataSourcePasswordReplacingFilter`, and `mcd` is another custom filter I've built.

When I run the Spark application the Spline agent initialization fails with:
```
21/12/21 17:32:15 ERROR QueryExecutionEventHandlerFactory: Spline initialization failed! Spark Lineage tracking is DISABLED.
java.util.NoSuchElementException: Missing configuration property spline.postProcessingFilter.mcd,password.className
```
Seems the comma in there isn't serving as a delimiter of filter names. I've tried a few different iterations of spacing around the comma with no effect. Any suggestions on how I can get this working in code instead of a config file?

_Originally posted by @97nitt in https://github.com/AbsaOSS/spline-spark-agent/discussions/369#discussioncomment-1855320_

97nitt:
Awesome. Thanks for the quick turnaround @wajda!

PR #371, opened 2021-12-22
issue #370 Spark Config :: Arrays properties don't work

wajda:
Make `SparkConfiguration` and `HadoopConfiguration` to behave as `MapConfiguration` with respect to multi-value properties treatment.

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/conf/HadoopConfiguration.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/conf/HadoopConfiguration.scala
@@ -16,11 +16,10 @@
 
 package za.co.absa.spline.harvester.conf
 
-import java.util
-
 import org.apache.hadoop.conf.{Configuration => SparkHadoopConf}
 
-import scala.collection.JavaConverters._
+import java.util
+import java.util.function.Consumer
 
 
 /**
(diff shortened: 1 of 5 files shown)


core/src/main/scala/za/co/absa/spline/harvester/conf/HadoopConfiguration.scala
```diff
@@ -16,25 +16,28 @@
...
+  extends ReadOnlyConfiguration
+    with MapLikeConfigurationAdapter {
+
+  override protected def propertiesMap: util.Map[String, _ <: AnyRef] = {
+    val aMap = new util.HashMap[String, String]()
+    shc.iterator().forEachRemaining(new Consumer[util.Map.Entry[String, String]] {
+      override def accept(t: util.Map.Entry[String, String]): Unit = {
+        aMap.put(t.getKey, t.getValue)
+      }
+    })
```
cerveada, line 40: Could a function be used instead of the Consumer?

cerveada (approved):

core/src/main/scala/za/co/absa/spline/harvester/conf/HadoopConfiguration.scala
wajda: Seems like lambda to functional interface automatic conversion only properly works since Scala 2.12

PR #377, opened 2021-12-28
issue #376 Kafka Dispatcher :: Producer API multi-version support

### LINKED ISSUE #376: Kafka Dispatcher :: Producer API multi-version support
Similarly to the `HttpLineageDispatcher`, `KafkaLineageDispatcher` should be able to send output in different Producer API versions. The required version number should be set in the dispatcher configuration defaulting to the current supported version (1.2).

wajda:
- Added `spline.lineageDispatcher.kafka.apiVersion` configuration property with the default value `1.2` that corresponds the latest Producer API version.
- Slightly reformatted the `spline.default.properties` to make it easier on eyes.
- Moved `ModelMapper` one level up in the package structure, so that it's reusable by other dispatchers
- Refactored `KafkaLineageDispatcher` and made it to use `ModelMapper` in a similar way the `HttpLineageDispatcher` does it.
- Added `ABSA-Spline-Entity-Type` header to represent the entity type.
- Changed semantics of the `__TypeId__` header to represent the entity _and_ the version combined (similar to the `Content-Type` header semantics in the HTTP protocol).
- Added some unit tests


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/KafkaLineageDispatcher.scala
```diff
@@ -17,85 +17,63 @@
...
-
   override def name = "Kafka"
 
-  private val planKafkaHeaders = Seq[Header](
-    new KafkaHeader(SplineKafkaHeaders.ApiVersion, ProducerApiVersion.Default.asString),
-    new KafkaHeader(SplineKafkaHeaders.TypeId, "ExecutionPlan")
-  )
+  private val senderFactory = {
+    val sf = createSenderFactory(apiVersion)
+    sys.addShutdownHook(sf.close())
```
cerveada, line 53: This seems weird. It looks like closing the factory. 

Also, the `AutoCloseable` there: I don't think a factory should really be handling closing of resources.

wajda: it actually _is_ closing a factory. The rational here is AMD cycle management - who created a resource is also responsible to dispose it properly. Hence the factory also has it's own lifecycle and here we ask the factory to shutdown.

cerveada: > who created a resource is also responsible to dispose it properly

The factory is creating a resource, but KafkaLineageDispatcher has to dispose it. Why not call `sys.addShutdownHook` inside the factory?

What do you mean by AMD?

wajda: > What do you mean by AMD?

Add, Modify, Delete

> Why not call `sys.addShutdownHook` inside the factory?

I didn't want a resource to outlive the factory. The idea here is to let the factory to manage the life cycle of the resource(s) it produces. So that from the user perspective, I call a factory to obtain a resource instance and just use it without having to think about if the returned resource is shared or isolated. All I need is to close a factory when I'm done and the factory will decide how to dispose its products properly.
Strictly speaking, I don't like `sys.addShutdownHook` here either as it basically leads to the same issue as if it were in the factory - the resource's and the factory's lifetime is longer that the dispatcher's one. Ideally I would want the dispatcher to have it's own life cycle management as well, but currently I don't see a good way of doing it.... apart maybe hooking the `SparkListener.onApplicationEnd()` event... The app still outlives the dispatcher and the session, but its lifecycle is shorter than the JVM one, thus it looks more appropriate hook than the JVM shutdown hook. What do you think? Since we have AgentBOM we can easily tie all agent components (dispatchers, filters etc) life cycle to the Spark context life cycle, and with that basically tie the entire Spline agent life cycle to it as well. So just in case several Spark contexts are started and stopped in one JVM the associated Spline agents would not leak any orphan resources.

cerveada: I think you are talking about other concern, `sys.addShutdownHook` is ok for this simple case I think.

My main point was that something named **factory** should be responsible solely for object creation and not it's destruction. Otherwise, it would be named **manager** or something else. Having the close method on the factory is making the code confusing.

Is the whole class really needed? It looks like a method in Dispatcher could do the same and would avoid this confusion completely.

cerveada (approved):

PR #413, opened 2022-03-07
spark agent #408 fix ObjectStructureDumper exception log bug

### LINKED ISSUE #408: ObjectStructureDumper - exception handling fails when stackTrace not present in Exception
As mentioned in #96

#### Other small changes needed for OSD
Don create dump of objects/instnaces of (toString is enough)
- `java.util.Random`
- `scala.util.Random`
- `java.lang.Number`
- `scala.math.Numeric`

cerveada:
- avoid exception when stack trace not present / null
- avoid infinite cycle when dumping Random
- additional types to print immediately


wajda (approved):
LGTM

PR #428, opened 2022-04-27
spark agent #427 catch LinkageErrors in ObjectStructureDumper

### LINKED ISSUE #427: Do not propagate LinkageError from ObjectStructureDumper
Even though LinkageError is considered fatal by scala, it should not be propagated from ObjectStructureDumper that serves just for logging.

Discoverd because of #426

cerveada:
fixes #427


core/src/main/scala/za/co/absa/spline/harvester/logging/ObjectStructureDumper.scala
```diff
@@ -96,6 +96,7 @@ object ObjectStructureDumper {
                 try {
                   extractFieldValue(value, f.getName)
                 } catch {
+                  case e: LinkageError => s"! error occurred: ${e.toShortString}"
                   case NonFatal(e) => s"! error occurred: ${e.toShortString}"
```
wajda: Just for the sake of brevity
```suggestion
                  case e@(_: LinkageError | NonFatal(_)) => s"! error occurred: ${e.toShortString}"
```

wajda (approved):

PR #439, opened 2022-06-08
spark agent #396 option for http dispatcher to set api version and co…

### LINKED ISSUE #396: HttpLineageDispatcher to not expect Spline server on the other end
From AbsaOSS/spline#1015 it might be useful if the user could configure `HttpLineageDispatcher` to not negotiate the protocol version and use a fixed one instead. 
For example (similar to `KafkaLineageDispatcher`):
```properties
# disables protocol version negotiation and sets the version to 1.2
spline.lineageDispatcher.http.apiVersion=1.2

# disables protocol version negotiation and sets the version to the latest supported one
spline.lineageDispatcher.http.apiVersion=LATEST
```

cerveada:
…mpression

fixes #396

### DIFF
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -36,10 +36,13 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
-# disables protocol version negotiation and sets the version to 1.2 ( LATEST could be used as well)
-# spline.lineageDispatcher.http.apiVersion=1.2
-# disables requestCompression negotiation and sets it to false
-# spline.lineageDispatcher.http.requestCompression=false
+# Spline Producer API version to use.
+# If unspecified, the protocol version is determined from the endpoint handshake response.
+# If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).
+# spline.lineageDispatcher.http.apiVersion=1 | 1.1 | 1.2 | LATEST
+# Should the payload be compressed.
+# If unspecified, the request compression is negotiated with the endpoint, otherwise it's disabled.
+# spline.lineageDispatcher.http.requestCompression=true | false
 
 # Kafka dispatcher
 spline.lineageDispatcher.kafka.className=za.co.absa.spline.harvester.dispatcher.KafkaLineageDispatcher
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -36,6 +36,12 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
+# set to false to use manual configuration of api version and compression
+spline.lineageDispatcher.http.configureUsingServer=true
+# the rest of http configs is used only when configureUsingServer=false
+# sets the version to LATEST ( 1.2 could be used as well)spark
+spline.lineageDispatcher.http.apiVersion=LATEST
+spline.lineageDispatcher.http.requestCompression=false
 
 # Kafka dispatcher
 spline.lineageDispatcher.kafka.className=za.co.absa.spline.harvester.dispatcher.KafkaLineageDispatcher
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -39,10 +39,10 @@ spline.lineageDispatcher.http.timeout.read=120000
 # Spline Producer API version to use.
 # If unspecified, the protocol version is determined from the endpoint handshake response.
 # If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).
-# spline.lineageDispatcher.http.apiVersion=1 | 1.1 | 1.2 | LATEST
+# spline.lineageDispatcher.http.api.version=1 | 1.1 | 1.2 | LATEST
 # Should the payload be compressed.
 # If unspecified, the request compression is negotiated with the endpoint, otherwise it's disabled.
-# spline.lineageDispatcher.http.requestCompression=true | false
+# spline.lineageDispatcher.http.request.compression=true | false
 
 # Kafka dispatcher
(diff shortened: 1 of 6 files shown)


core/src/main/resources/spline.default.properties
```diff
@@ -36,6 +36,12 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
+# set to false to use manual configuration of api version and compression
+spline.lineageDispatcher.http.configureUsingServer=true
```
wajda: Why do we need this property?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/HttpLineageDispatcherConfig.scala
```diff
@@ -34,4 +38,13 @@ class HttpLineageDispatcherConfig(config: Configuration) {
   val producerUrl: String = config.getRequiredString(ProducerUrlProperty)
   val connTimeout: Duration = config.getRequiredLong(ConnectionTimeoutMsKey).millis
   val readTimeout: Duration = config.getRequiredLong(ReadTimeoutMsKey).millis
+
+  val configureUsingServer: Boolean = config.getRequiredBoolean(ConfigureUsingServer)
+  def apiVersion: Version = stringToVersion(config.getRequiredString(ApiVersion))
```
wajda: If we make `apiVersion` and `requestCompression` optional (which they are), we could get rid of the `configureUsingServer` parameter. The logic would be simple - if the value is not provided by the user then try to negotiate one with the server.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
```diff
@@ -36,24 +36,29 @@ import scala.util.control.NonFatal
 /**
  * HttpLineageDispatcher is responsible for sending the lineage data to spline gateway through producer API
  */
-class HttpLineageDispatcher(restClient: RestClient)
+class HttpLineageDispatcher(restClient: RestClient, config: HttpLineageDispatcherConfig)
```
wajda: Since `restClient` object already incorporates some of the parameters specified in the `config`, passing both objects into the constructor creates an implicit ambiguity about the source of truth for those parameters. For example, from the constructor caller perspective, the `timeout` settings would appear as it has to be specified twice - in the dispatcher config, and the rest client. 
I would rather avoid passing config object together with other objects that are logically derived from it.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/HttpLineageDispatcherConfig.scala
cerveada: The goal here was make it more transparent and user friendly. Defining some optional properties to avoid server communication on init seems a bit hacky.

core/src/main/resources/spline.default.properties
```diff
@@ -36,12 +36,10 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
-# set to false to use manual configuration of api version and compression
-spline.lineageDispatcher.http.configureUsingServer=true
-# the rest of http configs is used only when configureUsingServer=false
-# sets the version to LATEST ( 1.2 could be used as well)spark
-spline.lineageDispatcher.http.apiVersion=LATEST
-spline.lineageDispatcher.http.requestCompression=false
+# disables protocol version negotiation and sets the version to 1.2 ( LATEST could be used as well)
```
wajda, line 39:
```suggestion
# Spline Producer API version. Available options: 1, 1.1, 1.2, LATEST
# If unspecified, the protocol version is determined from the endpoint handshake response.
# If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).
```

core/src/main/resources/spline.default.properties
```diff
@@ -36,12 +36,10 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
...
+# spline.lineageDispatcher.http.apiVersion=1.2
+# disables requestCompression negotiation and sets it to false
```
wajda, line 41:
```suggestion
# Should the payload be compressed.
# If unspecified, the request compression is negotiated with the endpoint, otherwise it's disabled.
```

wajda (dismissed):
See my suggestion for property documentation. Otherwise LGTM.

core/src/main/resources/spline.default.properties
```diff
@@ -36,6 +36,10 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
+# disables protocol version negotiation and sets the version to 1.2 ( LATEST could be used as well)
+# spline.lineageDispatcher.http.apiVersion=1.2
```
wajda:
```suggestion
# Spline Producer API version to use.
# If unspecified, the protocol version is determined from the endpoint handshake response.
# If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).
# spline.lineageDispatcher.http.apiVersion=1 | 1.1 | 1.2 | LATEST
```

core/src/main/resources/spline.default.properties
```diff
@@ -36,6 +36,10 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
...
+# disables requestCompression negotiation and sets it to false
+# spline.lineageDispatcher.http.requestCompression=false
```
wajda:
```suggestion
# Should the payload be compressed.
# If unspecified, the request compression is negotiated with the endpoint, otherwise it's disabled.
# spline.lineageDispatcher.http.requestCompression=true | false
```

core/src/main/resources/spline.default.properties
cerveada: > If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).

they still need the status endpoint to respond `200 OK` as it was before, but when implementing endpoints for plan and events this should be easy to add as well.

wajda: By that I meant that custom endpoints would hardly send any Spline custom headers, and if one isn't sent, the version defaults to 1

PR #444, opened 2022-06-10
Feature/agent 443 fix compiler warnings

cerveada:


core/src/main/scala/za/co/absa/spline/harvester/idGenerators.scala
```diff
@@ -66,15 +63,22 @@ class UUID3IdGenerator[-A <: AnyRef](namespace: UUID) extends HashBasedUUIDGener
...
-  def nextId(unused: Any): String = {
+  def nextId(): String = {
     val i = counter.getAndIncrement()
     MessageFormat.format(pattern, Long.box(i))
   }
 }
+
+class DataTypeIdGenerator(numberTemplate: String, namespace: UUID) {
+  val seqIdGen = new SequentialIdGenerator(numberTemplate)
+  val UUIDGen = new UUID5IdGenerator[String](namespace)
```
wajda:
```suggestion
  val uuidGen = new UUID5IdGenerator[String](namespace)
```
it's a local variable, so lowercase seems more appropriate

wajda: otherwise LGTM

PR #447, opened 2022-06-14
spark agent #53 add fallback lineage dispatcher

### LINKED ISSUE #53: Add fallback option in case Producer is down and write doesn't pass
If in case of reasons unknown spline is not able to log the lineage to the ArangoDB, I would love to have a fallback option of writing the data to local disk or HDFS, which then I would be able to manually insert into the ArangoDB. 

### Why?
I might be allowed to only run the job once or once every X hours/days/weeks which means my data about the job is lost.

### Additions requests:
- I would not like to lose the default configuration that works almost flawlessly out of the box
- the path to the file should be configurable by the end-user
- if it could be a pre-prepared ArangoDB script that I just run

cerveada:
fixes #53


wajda (approved):
LGTM

PR #475, opened 2022-07-18
sparkUser added

### LINKED ISSUE #474: Capture 'sparkUser' in the event extras
as per #473

Since it's available in the `SparkContext`, we can easily capture it by default. Many people ask for that.
Perhaps we could simply create a predefined rules for the `userExtraMeta` filter and use it as a default.

vishalag001:
Added sparkUser to extras to plan and event Object

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
@@ -83,6 +83,7 @@ class LineageHarvester(
         val planWithoutId = {
           val planExtra = Map[String, Any](
             ExecutionPlanExtra.AppName -> ctx.session.sparkContext.appName,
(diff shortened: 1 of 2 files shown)


vishalag001:
@wajda Please review the above PR

wajda:
Thank you for the PR.
My only question would be why do you capture the user in both execution plan and event objects?
I think this information is only relevant for the execution event, because the plan record can be linked to multiple runs (events) potentially triggered by different users.

wajda:
Also, can you please change the commit message, so that it refers the issue number. We have a rule to always refer the issue number in commit messages. Something like `Issue #474 capture sparkUser`

vishalag001:
I'll remove the user info from plan and include the issue number too

vishalag001:
@wajda I have updated the PR

wajda (approved):
LGTM

wajda:
Merged. Thank you @vishalag001 for contribution.

PR #493, opened 2022-09-02
Feature/agent 33 rdd support

### LINKED ISSUE #498: RDD lineage support (for RDD wrapped in LogicalPlan)
Support lineage of this type:
```
WRITE(LogicalPlan) -- OP(LogicalPlan) -- OP(RDD) -- READ(RDD)
```

This type of lineage is possible to capture, since the Write triggers Spark's `QueryExecutionListener` tha Spline already uses to capture the LogicalPlan lineage.

cerveada:
This PR adds rdd support for cases when the write itself is not rdd but somewhere in the lineage graph an rdd is used. So it captures operation in the middle and read operation.

- add another builder branch for processing RDDs
- refactor builders for plan and rdd into separated sub-packages
- add another plugin type for RDD reads
- rename variables to keep distinction between Spark's plan and Spline Operation abstraction
- add leaf finding capabilities to LineageWalker
- add test for RDD lineages

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
@@ -19,9 +19,10 @@ package za.co.absa.spline.harvester
 import org.apache.commons.lang3.exception.ExceptionUtils
 import org.apache.spark
 import org.apache.spark.internal.Logging
+import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SaveMode
 import org.apache.spark.sql.catalyst.plans.logical._
-import org.apache.spark.sql.execution.{LeafExecNode, SparkPlan}
+import org.apache.spark.sql.execution.{ExternalRDD, LeafExecNode, LogicalRDD, SparkPlan}
 import za.co.absa.commons.CollectionImplicits._
 import za.co.absa.commons.graph.GraphImplicits._
 import za.co.absa.commons.lang.CachingConverter
@@ -158,8 +159,8 @@ class LineageHarvester(
     @scala.annotation.tailrec
     def traverseAndCollect(
       accBuilders: Seq[OperationNodeBuilder],
-      processedEntries: Map[LogicalPlan, OperationNodeBuilder],
-      enqueuedEntries: Seq[(LogicalPlan, OperationNodeBuilder)]
+      processedEntries: Map[PlanOrRdd, OperationNodeBuilder],
+      enqueuedEntries: Seq[(PlanOrRdd, OperationNodeBuilder)]
     ): Seq[OperationNodeBuilder] = {
       enqueuedEntries match {
         case Nil => accBuilders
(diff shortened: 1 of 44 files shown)


core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
```diff
@@ -184,29 +185,41 @@ class LineageHarvester(
...
   }
 }
 
 object LineageHarvester {
 
   import za.co.absa.commons.version.Version
 
+  trait PlanOrRdd
+  case class PlanWrap(plan: LogicalPlan) extends PlanOrRdd
+  case class RddWrap(rdd: RDD[_]) extends PlanOrRdd
```
wajda, line 221: Why didn't you want to use `Either`? I assume readability was one of the reasons, but I wonder if there's something else?

integration-tests/src/main/scala/za/co/absa/spline/test/ProducerModelImplicits.scala
```diff
@@ -28,20 +30,33 @@ object ProducerModelImplicits {
...
   }
 
   implicit class WriteOperationOps(val write: WriteOperation) extends AnyVal {
 
-    def precedingOp(implicit walker: LineageWalker): DataOperation =
-      walker.precedingOp(write)
+    def precedingDataOp(implicit walker: LineageWalker): DataOperation = {
+      val children = walker.precedingDataOps(write)
+      assert(children.size == 1)
+      children.head
```
wajda: Here I would use a _destructuring assignment_ syntax as I think it communicates the intention better than the `assert` statement. Also because `assert` call could  be skipped at runtime nothing would prevent the rest of the values from being silently skipped should they happen to appear in the list for whatever reason.
```suggestion
      val Seq(child) = walker.precedingDataOps(write)
      child
```

integration-tests/src/main/scala/za/co/absa/spline/test/ProducerModelImplicits.scala
```diff
@@ -53,4 +68,34 @@ object ProducerModelImplicits {
     }
   }
 
+  private def findLeaves(operation: Operation)(implicit walker: LineageWalker): (Seq[ReadOperation], Seq[DataOperation]) = {
+
+    @tailrec
+    def findLeavesRec(
+      toVisit: List[Operation],
+      readLeaves: List[ReadOperation],
+      dataLeaves: List[DataOperation]
```
wajda, line 75: Why do you `List` instead of `Seq` here?

cerveada: - List syntax is nice match for this type of algorithm with lot of head tail operations, usable and readable
- List as data structure also works well for what I am trying to do
- Don't see a reason not to

core/src/main/scala/za/co/absa/spline/harvester/LineageHarvester.scala
cerveada: Readability: yes, also `Either` I see as more suitable for cases when Left represents some failed state.

From docs:
> Either is right-biased, which means that Right is assumed to be the default case to operate on. If it is Left, operations like map and flatMap return the Left value unchanged

So, thats not what I want here, both values are equally correct in this case.

integration-tests/src/main/scala/za/co/absa/spline/test/ProducerModelImplicits.scala
wajda: What has caught my eye is the `toList` call on a `Seq` in two places.
I understand your intention, and of course I likely am being nitpicking here as in reality those collections in 90-ish % of cases are too small for their type to make any perceivable impact on performance. But since we are talking about it judging from the given method code it doesn't look persuading that you are benefiting from the `List` collection when dealing with the `toVisit` variable, as it in fact might be anything prior converting to a `List`. So every time I see the collection type conversion the question "Why?" pops up in my head. You never know the cost of conversion of a more generic collection type to a more specific one. So to me the rule of thumb is that there always have to be really good reasons to do so.

I agree about the rest two variables though.

cerveada: But isn't by the same logic potential cost of head, tail or append on an unknown collection also problematic? The conversion happens just once for each children collection, but the other operations are called for each element in that collection, so multiple times.

wajda: Well, according to [this](https://docs.scala-lang.org/overviews/collections/performance-characteristics.html) the majority standard immutable collections execute `head` and `tail` in a constant (or effectively constant) time, while conversion (that most likely requires iteration) will very probably be `O(n)`. The complexity of the whole function seems to be linear as well. To me that means that you would hardly make the overall performance worse if you leave generic `Seq`. And if you do convert, although you won't make it worth either from the `O` perspective, but you may likely increase a constant factor of `N` of `findLeavesRec`'s complexity,. For what it's worth, of course.

integration-tests/src/main/scala/za/co/absa/spline/test/ProducerModelImplicits.scala
cerveada: fixed

integration-tests/src/main/scala/za/co/absa/spline/test/ProducerModelImplicits.scala
cerveada: changed it to Seq

wajda (approved):
LGTM

dk1844 (approved):
LGTM (just read the code)

PR #495, opened 2022-09-20
Adding security header for spline http dispatcher

rupesh3020:
As per discussion in https://github.com/AbsaOSS/spline-spark-agent/discussions/448
A security property can be added in http property in spline.properties

### DIFF
--- a/core/src/main/resources/spline.default.properties
+++ b/core/src/main/resources/spline.default.properties
@@ -52,6 +52,7 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
+spline.lineageDispatcher.http.securityHeader.key=somekeys
 # Spline Producer API version to use.
 # If unspecified, the protocol version is determined from the endpoint handshake response.
 # If unspecified and a custom (non-Spline) endpoint is used, the Producer API version is automatically set to 1 (the oldest one supported).
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/SplineHeaders.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/SplineHeaders.scala
@@ -26,6 +26,7 @@ object SplineHeaders {
   // Http specific
   val AcceptRequestEncoding = s"$Prefix-Accept-Request-Encoding"
   val Timeout = "X-SPLINE-TIMEOUT"
+  val HttpSecurityKey = "X-FUNCTIONS_KEY"
 
   // Kafka specific
   val SplineEntityType = s"$Prefix-Entity-Type"
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcher.scala
@@ -107,7 +107,8 @@ object HttpLineageDispatcher extends Logging {
       Http,
       config.producerUrl,
       config.connTimeout,
-      config.readTimeout
+      config.readTimeout,
+      config.secHeader
     )
   }
 
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/HttpLineageDispatcherConfig.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/HttpLineageDispatcherConfig.scala
@@ -30,6 +30,7 @@ object HttpLineageDispatcherConfig {
   val ReadTimeoutMsKey = "timeout.read"
   val ApiVersion = "apiVersion"
   val RequestCompression = "requestCompression"
+  val securityHeaderKey = "securityHeader.key"
 
   def apply(c: Configuration) = new HttpLineageDispatcherConfig(c)
 }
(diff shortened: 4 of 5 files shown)


core/src/main/resources/spline.default.properties
```diff
@@ -52,6 +52,7 @@ spline.lineageDispatcher.http.className=za.co.absa.spline.harvester.dispatcher.H
 spline.lineageDispatcher.http.producer.url=
 spline.lineageDispatcher.http.timeout.connection=2000
 spline.lineageDispatcher.http.timeout.read=120000
+spline.lineageDispatcher.http.securityHeader.key=somekeys
```
cerveada, line 55: the security headers should not be mandatory, but optional (as apiVersion and requestCompression is)

also is probably better to not provide any default value here

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/SplineHeaders.scala
```diff
@@ -26,6 +26,7 @@ object SplineHeaders {
   // Http specific
   val AcceptRequestEncoding = s"$Prefix-Accept-Request-Encoding"
   val Timeout = "X-SPLINE-TIMEOUT"
+  val HttpSecurityKey = "X-FUNCTIONS_KEY"
```
cerveada, line 29: as I understand it X-FUNCTIONS_KEY is Azure specific header. It doesn't make sense to add one specialised header like that. Why not implement it in a away that the user can add any header they want?

cerveada: One idea to do it would be to have something like this:
```
spline.lineageDispatcher.http.header.customName.key=X-FUNCTIONS_KEY
spline.lineageDispatcher.http.header.customName.value=somekeys

spline.lineageDispatcher.http.header.customName2.key=SecondHeaderKey
spline.lineageDispatcher.http.header.customName2.value=SecondHeaderValue
```
So the user could define as many headers as they want. `customName`'s function in this is just to pair the key and the value.

rupesh3020: Yes , i agree with you. thanks for the idea.

core/src/main/resources/spline.default.properties
rupesh3020: I wanted the same but i am not sure how we can pass the n number of header as request compression and apiversion is added as argument to HTTPDispather class.

rupesh3020: Is it ok to create a seperate HttpLineageDispather for Azure functions like for open lineage is done. I see the openLineageHttpDispatcher is dependent on RestClient from Httpdispather. This will solve my team problem and will also not be a big change

cerveada: > I wanted the same but i am not sure how we can pass the n number of header as request compression and apiversion is added as argument to HTTPDispather class.

Assuming the config is in this format:
```
spline.lineageDispatcher.http.header.customName.key=X-FUNCTIONS_KEY
spline.lineageDispatcher.http.header.customName.value=somekeys

spline.lineageDispatcher.http.header.customName2.key=SecondHeaderKey
spline.lineageDispatcher.http.header.customName2.value=SecondHeaderValue
```
You can do the extraction in `HttpLineageDispatcherConfig` class like : 
```
val headersConfig = config.subset("header")
```
then you can do `headersConfig.getKeys` and for each header get the key and value and put them in the scala map. That map can be a dispatcher parameter. `Map[String, String]`

cerveada: > Is it ok to create a seperate HttpLineageDispather for Azure functions like for open lineage is done. I see the openLineageHttpDispatcher is dependent on RestClient from Httpdispather. This will solve my team problem and will also not be a big change

It's ok to create as many dispatchers as you want, but if you want them to become part of Spline, they must be generic enough to be useful for others.

If only difference is one header it means there is lot of functionality duplicated.

If you see there is a lot of things needed for Azure that are different from the standard dispatcher it would make sense to do AzureDispatcher, but in case of adding headers I think it would be useful to have them in generic http dispatcher anyway.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/SplineHeaders.scala
wajda: I would simplify the config a little bit by combining `key` and `value` properties into one `key=value` and removing the `customName` part.
For example:
```properties
spline.lineageDispatcher.http.header.X-FUNCTIONS_KEY=my-function-key
```

cerveada: > I would simplify the config a little bit by combining `key` and `value` properties into one `key=value` 

I was thinking about that as well, but I thought there might be cases when the header is not valid property key. But I am not aware of any such headers, so maybe I am just too paranoid here.

wajda: RFC 9110, section [16.3.2.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-considerations-for-new-field):
> While the field-name syntax is defined to allow any token character, in practice some implementations place limits on the characters they accept in field-names. **To be interoperable, new field names SHOULD constrain themselves to alphanumeric characters, "-", and ".", and SHOULD begin with a letter**. For example, the underscore ("_") character can be problematic when passed through non-HTTP gateway interfaces (see [Section 17.10](https://www.rfc-editor.org/rfc/rfc9110.html#underscore.in.fields)).

So I guess we are pretty much safe here :)

cerveada:
@rupesh3020 since you solved your issue other way I implemented the more generic solution I suggested in separated PR #500

wajda:
Created a separate ticket for Azure specific dispatcher config - #501

PR #541, opened 2022-11-25
#536 - tests fails on windows

### LINKED ISSUE #536: Tests fail on Windows
Having tried to run `tests` on Windows (Win 11 x64) with docker-desktop alternative running, I am failing to test-run the project.

I have run `mvn test > out.txt`, attaching a log
[out.zip](https://github.com/AbsaOSS/spline-spark-agent/files/10075978/out.zip)

@cerveada @miroslavpojer

miroslavpojer:
- fixed 12 failing tests related to windows-like paths comparison

Closes #536

### DIFF
--- a/integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
+++ b/integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
@@ -34,6 +34,7 @@ trait SparkFixture {
   val warehouseDir: String = baseDir.asString.stripSuffix("/")
   val warehouseUri: URI = baseDir.toURI
   val metastoreDir: String = TempDirectory("Metastore", "debug", pathOnly = true).deleteOnExit().asString
+  val metastoreConnectURL = s"jdbc:derby:$metastoreDir/metastore_db;create=true"
 
   protected val sessionBuilder: SparkSession.Builder = {
     SparkSession.builder
@@ -41,7 +42,8 @@ trait SparkFixture {
       .config("spark.driver.host","localhost")
       .config("spark.sql.warehouse.dir", warehouseDir)
       .config("spark.ui.enabled", "false")
-      .config("javax.jdo.option.ConnectionURL", s"jdbc:derby:$metastoreDir/metastore_db;create=true")
+      .config("javax.jdo.option.ConnectionURL", metastoreConnectURL)
+      .config("hive.metastore.warehouse.dir", metastoreConnectURL) // required by Spark 2.3
   }
 
   def withSparkSession[T](testBody: SparkSession => T): T = {
--- a/integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
+++ b/integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
@@ -22,6 +22,7 @@ import org.scalatest.{Assertion, AsyncTestSuite, BeforeAndAfterEach}
 import za.co.absa.commons.io.TempDirectory
 
 import java.io.File
+import java.net.URI
 import java.sql.DriverManager
 import scala.concurrent.Future
 import scala.util.Try
@@ -30,8 +31,9 @@ trait SparkFixture {
   this: AsyncTestSuite =>
 
   val baseDir: TempDirectory = TempDirectory("SparkFixture", "UnitTest", pathOnly = true).deleteOnExit()
-  val warehouseDir: String = baseDir.path.toAbsolutePath.toString.stripSuffix("/")
-  val warehouseUri: String = new File(warehouseDir).toURI.toString
+  val warehouseDir: String = baseDir.asString.stripSuffix("/")
+  val warehouseUri: URI = baseDir.asURI
+  val metastoreDir: String = TempDirectory("Metastore", "debug", pathOnly = true).deleteOnExit().asString
 
   protected val sessionBuilder: SparkSession.Builder = {
     SparkSession.builder
@@ -39,6 +41,7 @@ trait SparkFixture {
       .config("spark.driver.host","localhost")
       .config("spark.sql.warehouse.dir", warehouseDir)
       .config("spark.ui.enabled", "false")
+      .config("javax.jdo.option.ConnectionURL", s"jdbc:derby:$metastoreDir/metastore_db;create=true")
   }
 
   def withSparkSession[T](testBody: SparkSession => T): T = {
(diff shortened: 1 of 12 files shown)


wajda:
Now tests are failing on Linux
```
org.scalatest.exceptions.TestFailedException: URI file:/tmp/SparkFixture3286681055411943221UnitTest/sourcetable was not the same as expected file://tmp/SparkFixture3286681055411943221UnitTest/sourcetable
```

miroslavpojer:
> Now tests are failing on Linux
> 
> ```
> org.scalatest.exceptions.TestFailedException: URI file:/tmp/SparkFixture3286681055411943221UnitTest/sourcetable was not the same as expected file://tmp/SparkFixture3286681055411943221UnitTest/sourcetable
> ```

Solved in following commits. Build server has been used to run & show these problems as I am running on windows machine.

integration-tests/src/test/scala/za/co/absa/spline/ExcelSpec.scala
```diff
@@ -25,12 +25,15 @@ import za.co.absa.commons.io.{TempDirectory, TempFile}
...
 
+import java.io.File
+
 class ExcelSpec extends AsyncFlatSpec
   with Matchers
   with SparkFixture
   with SplineFixture {
 
   private val filePath = TempFile("file1", ".xlsx", false).deleteOnExit().path.toAbsolutePath.toString
+  private val fileUri = new File(filePath).toURI.toString
```
wajda: minor:
unnecessary `toString` call. The name `fileUri` suggests it is of the `URI` type, while in fact is `String`. If would be OK in other situations where it's passed to the expressions where a string is expected. But in this particular context (and the other similar ones in this PR) the given variable is only used in string interpolation, where `Any` is expected and `toString` is called implicitly. So it would be more natural for `fileUri` variable to be of the `URI` type.

integration-tests/src/test/scala/za/co/absa/spline/ExcelSpec.scala
```diff
@@ -63,7 +66,7 @@ class ExcelSpec extends AsyncFlatSpec
         } yield {
           plan1.operations.write.append shouldBe false
           plan1.operations.write.extra.get("destinationType") shouldBe Some("excel")
-          plan1.operations.write.outputSource shouldBe s"file:$filePath"
+          plan1.operations.write.outputSource shouldBe s"$fileUri"
```
wajda: Redundant string interpolation:
```suggestion
          plan1.operations.write.outputSource shouldBe fileUri
```

wajda (commented):
LGTM excepts for redundant `toString` and `s"$xxx"` as commented below.

integration-tests/src/test/scala/za/co/absa/spline/ExcelSpec.scala
cerveada: maybe `toURI` method should be added to `TempFile` and `TempDirectory` so you can call 
```scala
private val fileUri = TempFile("file1", ".xlsx", false).deleteOnExit().toURI
```

integration-tests/src/test/scala/za/co/absa/spline/SQLCommandsSpec.scala
```diff
@@ -58,9 +60,9 @@ class SQLCommandsSpec extends AsyncFlatSpec
                   | WHERE id > 1""".stripMargin))
 
           } yield {
-            plan1.operations.write.outputSource should equalToUri(s"file:$warehouseDir/sourcetable")
+            plan1.operations.write.outputSource should equalToUri(s"${warehouseUri.toString}/sourcetable")
             plan2.operations.reads.get.head.inputSources.head should equalToUri(plan1.operations.write.outputSource)
-            plan2.operations.write.outputSource should equalToUri(s"file:$warehouseDir/targettable")
+            plan2.operations.write.outputSource should equalToUri(s"${warehouseUri.toString}/targettable")
```
cerveada: You don't need `toString` here it's called automatically when using `s"$warehouseUri"`

integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
```diff
@@ -52,7 +53,7 @@ class LineageHarvesterSpec extends AsyncFlatSpec
       withLineageTracking { captor =>
         import spark.implicits._
 
-        val testDest = tmpDest
+        val testDest = tmpDest.asString
```
cerveada: In this case the code would be clearer using `tmpDest()` to make it known that method is called. I thought it was variable for some time.

Also it is recomended in scala style guide
https://docs.scala-lang.org/style/method-invocation.html#arity-0

cerveada: or you could just delete the method and use the code directly it's really short and it would make it obvious what is happening.

integration-tests/src/test/scala/za/co/absa/spline/ExcelSpec.scala
miroslavpojer: Agreed. Solved in following series of commit.

integration-tests/src/test/scala/za/co/absa/spline/SQLCommandsSpec.scala
miroslavpojer: Fixed in commit e54f5ef9a17cf0c9ad82fbc7081d401b31a933b3.

integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
miroslavpojer: Fixed in commit 5d0f9ae49f2b674bea77d05b7807d51df7eacc06.

integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
```diff
@@ -22,21 +22,27 @@ import org.scalatest.{Assertion, AsyncTestSuite, BeforeAndAfterEach}
...
+  val metastoreDir: String = TempDirectory("Metastore", "debug", pathOnly = true).deleteOnExit().asString
+  val metastoreConnectURL = s"jdbc:derby:$metastoreDir/metastore_db;create=true"
 
   protected val sessionBuilder: SparkSession.Builder = {
     SparkSession.builder
       .master("local")
       .config("spark.driver.host","localhost")
       .config("spark.sql.warehouse.dir", warehouseDir)
       .config("spark.ui.enabled", "false")
+      .config("javax.jdo.option.ConnectionURL", metastoreConnectURL)
```
cerveada, line 45: What is this config needed for?

cerveada (approved):

wajda (approved):
lgtm

integration-tests/src/main/scala/za/co/absa/spline/test/fixture/SparkFixture.scala
miroslavpojer: It is used to define custom paths to metastore_db. It solves an issue found on windows where default location was still hold by process when trying to apply clean logic.

PR #544, opened 2022-11-25
spark agent #517 delta - delete, update, merge into support

### LINKED ISSUE #517: Support for DELETE, UPDATE and MERGE in Delta
Hi,

We are using following code for capturing lineage
```python
from delta.tables import *
from pyspark.sql.functions import *

oldIncrementalData = spark.range(6).withColumn("name", lit("Dhruv"))
old_deltaTable = DeltaTable.forPath(spark, "/data/events_old/")

newIncrementalData = spark.range(5).withColumn("name", lit("Neha"))
```
# Create DeltaTable instances using the path of the Delta table
```python
deltaTable = DeltaTable.forPath(spark, "/data/events/")
```
# Execute Upsert
```python
(old_deltaTable
 .alias("oldData") 
 .merge(newIncrementalData.alias("newData"), "oldData.id = newData.id")
 .whenMatchedUpdate(set = {"name": col("newData.name")})
 .whenNotMatchedInsert(values = {"id": col("newData.id"), "name": col("newData.name")})
 .execute()
)
```

Here SplineQueryExecutionListener is not getting called for merge operation.
Can you please help for the same.
Thanks

cerveada:
fixes #517

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DeltaPlugin.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DeltaPlugin.scala
@@ -34,10 +34,10 @@ import scala.language.reflectiveCalls
 @Priority(Precedence.Normal)
 class DeltaPlugin(
   pathQualifier: PathQualifier,
-  session: SparkSession)
-  extends Plugin
+  session: SparkSession
+) extends Plugin
     with WriteNodeProcessing
-    with ReadNodeProcessing{
+    with ReadNodeProcessing {
 
   override val writeNodeProcessor: PartialFunction[LogicalPlan, WriteNodeInfo] = {
     case `_: DeleteCommand`(command) =>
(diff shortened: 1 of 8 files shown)


core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DeltaPlugin.scala
```diff
@@ -75,7 +75,7 @@ class DeltaPlugin(
       val syntheticSourceRead = SyntheticDeltaRead(sourceRelation.output, sourceId, Map.empty[String, Any], sourceRelation)
       val params = Map.empty[String, Any]
 
-      val condition = extractValue[Expression](command, "condition").toString
+      val condition = extractValue[Option[Expression]](command, "condition").toString
```
wajda: Is `Option.toString` intended here? Shouldn't it be `.map(_.toString)`?

integration-tests/src/test/scala/za/co/absa/spline/DeltaDSV2Spec.scala
```diff
@@ -345,7 +345,7 @@ class DeltaDSV2Spec extends AsyncFlatSpec
             plan.operations.write.outputSource shouldBe s"file:$warehouseDir/testdb.db/foo"
 
             val mergeOp = plan.operations.write.precedingDataOp
-            mergeOp.params.get("condition").asInstanceOf[String] should include("ID")
+            mergeOp.params.get("condition").asInstanceOf[Option[String]].get should include("ID")
```
wajda: I guess it's better to use ScalaTest's `OptionalValues` implicits for matching options.
https://www.scalatest.org/user_guide/using_OptionValues

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DeltaPlugin.scala
cerveada: In DELETE and UPDATE case the params are set directly so I wanted to avoid going through OperationParamsConverter in this case only.

wajda (approved):
lgtm

PR #554, opened 2022-12-07
Feature/agent 551 remove options from model

### LINKED ISSUE #551: Remove meaningless Options from agent's data model

cerveada:


wajda (dismissed):

cerveada:
I removed the Options from everywhere where it made sense. I kept them only in several places where the values are actually optional. That is: `dataType` and `discriminator` and `id` on `ExecutionPlan`.

There is still a question of the mapping. `None` values will not appear in json at all, but `Some(Seq.empty)` will appear as empty json object `variableName: {}` . I think we should pick one of those approaches and stick to it. But I am not sure which is better. What do you think @wajda ?

wajda (dismissed):
LGTM (read the code, not tested)

PR #561, opened 2022-12-19
Feature/spline spark agent 559 yaml

### LINKED ISSUE #559: Support config in YAML format
Since Spline configuration is getting more and more complex and flexible expressing certain parts of it in Java properties can be cumbersome and tricky, for example encoding regexps or JSON pieces while keeping all escaping correct...
YAML would do much better in those cases.

wajda:
fixes #559


cerveada (dismissed):

core/src/main/scala/za/co/absa/spline/harvester/conf/StandardSplineConfigurationStack.scala
```diff
@@ -37,17 +37,19 @@ import scala.util.Try
  *
  */
 object StandardSplineConfigurationStack {
-  private val PropertiesFileName = "spline.properties"
-  private val DefaultPropertiesFileName = "spline.default.properties"
+  private val PropertiesFileName = "/spline.properties"
```
wajda:
```suggestion
  private val PropertiesFileName = "spline.properties"
```

cerveada (approved):

PR #582, opened 2023-01-25
Bugfix/spline spark agent 581 integration tests

### LINKED ISSUE #581: Core jar to depend on a fixed Spark version regardless of active Maven profile
`spark-xy` profile should only affect integration tests, not the _core_ jar.
The final production artifacts are always built with the fixed Spark version (2.4), and integration tests should test them as they are. Currently the Spark version is managed by `spark-x.y.` profiles for the entire project. Thus the Spark versions used for the _core_ module compilation and for the integration tests run are always properly aligned, which doesn't reflect the real usage.

wajda:
fixes #581 

1. Add ability to enable/disable any plugin in the configuration
2. Fix spark dependencies accordingly to the issue description

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/plugin/registry/AutoDiscoveryPluginRegistry.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/plugin/registry/AutoDiscoveryPluginRegistry.scala
@@ -23,7 +23,7 @@ import org.apache.spark.internal.Logging
 import za.co.absa.commons.lang.ARM
 import za.co.absa.spline.harvester.plugin.Plugin
 import za.co.absa.spline.harvester.plugin.Plugin.Precedence
-import za.co.absa.spline.harvester.plugin.registry.AutoDiscoveryPluginRegistry.{PluginClasses, getOnlyOrThrow}
+import za.co.absa.spline.harvester.plugin.registry.AutoDiscoveryPluginRegistry.{EnabledByDefault, EnabledConfProperty, PluginClasses, getOnlyOrThrow}
 
 import javax.annotation.Priority
 import scala.collection.JavaConverters._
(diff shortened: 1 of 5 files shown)


core/src/main/scala/za/co/absa/spline/harvester/plugin/registry/AutoDiscoveryPluginRegistry.scala
```diff
@@ -73,10 +73,22 @@ class AutoDiscoveryPluginRegistry(
     constr.newInstance(args: _*).asInstanceOf[Plugin]
   }
 
+  private def isPluginEnabled(pc: Class[Plugin]): Boolean = {
+    val pluginConf = conf.subset(pc.getName)
+    val isEnabled = pluginConf.getBoolean(EnabledConfProperty, EnabledByDefault)
+    if (!isEnabled) {
+      log.warn("Plugin {} is disabled in the configuration.", pc.getName)
```
cerveada: The message is evaluated lazily already no need for `{}` or is there?

wajda: it's SLF4J. Don't you confuse it with SLF4S? SLF4J is Java API, there's no pass-by-name for method params.

cerveada:
```
  protected def logWarning(msg: => String): Unit = {
    if (log.isWarnEnabled) log.warn(msg)
  }
```

cerveada:
```suggestion
      logWarning(s"Plugin ${pc.getName} is disabled in the configuration.")
```

wajda: Ah, Ok, I can use another method `logWarning` that is Scala one with pass-by-name :)

wajda: done

cerveada (dismissed):
LGTM

PR #606, opened 2023-02-16
spline #605 MergeIntoNodeBuilder: java.lang.IllegalArgumentException

### LINKED ISSUE #605: MergeIntoNodeBuilder: java.lang.IllegalArgumentException: transpose
```
java.lang.IllegalArgumentException: transpose requires all collections have the same size
	at scala.collection.generic.GenericTraversableTemplate.fail$1(GenericTraversableTemplate.scala:215)
	at scala.collection.generic.GenericTraversableTemplate.$anonfun$transpose$3(GenericTraversableTemplate.scala:222)
	at scala.collection.generic.GenericTraversableTemplate.$anonfun$transpose$3$adapted(GenericTraversableTemplate.scala:221)
	at scala.collection.immutable.List.foreach(List.scala:431)
	at scala.collection.generic.GenericTraversableTemplate.$anonfun$transpose$2(GenericTraversableTemplate.scala:221)
	at scala.collection.generic.GenericTraversableTemplate.$anonfun$transpose$2$adapted(GenericTraversableTemplate.scala:219)
	at scala.collection.immutable.List.foreach(List.scala:431)
	at scala.collection.generic.GenericTraversableTemplate.transpose(GenericTraversableTemplate.scala:219)
	at scala.collection.generic.GenericTraversableTemplate.transpose$(GenericTraversableTemplate.scala:211)
	at scala.collection.AbstractTraversable.transpose(Traversable.scala:108)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.mergeInputs$lzycompute(MergeIntoNodeBuilder.scala:32)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.mergeInputs(MergeIntoNodeBuilder.scala:32)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.outputAttributes$lzycompute(MergeIntoNodeBuilder.scala:37)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.outputAttributes(MergeIntoNodeBuilder.scala:36)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.build(MergeIntoNodeBuilder.scala:61)
	at za.co.absa.spline.harvester.builder.plan.MergeIntoNodeBuilder.build(MergeIntoNodeBuilder.scala:27)
	at za.co.absa.spline.harvester.LineageHarvester.$anonfun$harvest$6(LineageHarvester.scala:72)
	at scala.collection.immutable.List.map(List.scala:297)
	at za.co.absa.spline.harvester.LineageHarvester.$anonfun$harvest$4(LineageHarvester.scala:72)
	at scala.Option.flatMap(Option.scala:271)
	at za.co.absa.spline.harvester.LineageHarvester.harvest(LineageHarvester.scala:65)
	at za.co.absa.spline.agent.SplineAgent$$anon$1.$anonfun$handle$1(SplineAgent.scala:91)
	at za.co.absa.spline.agent.SplineAgent$$anon$1.withErrorHandling(SplineAgent.scala:100)
	at za.co.absa.spline.agent.SplineAgent$$anon$1.handle(SplineAgent.scala:72)
	at za.co.absa.spline.harvester.listener.QueryExecutionListenerDelegate.onSuccess(QueryExecutionListenerDelegate.scala:28)
```

_Originally posted by @wajda in https://github.com/AbsaOSS/spline-spark-agent/issues/570#issuecomment-1433075188_

wajda:
fixes #605 

- In the `MergeIntoNodeBuilder` replaced `inputAttribute.transpose` with a mapping algorithm based on attribute names. Transpose cannot be used for that purpose in this place as MERGE command can have inputs with different number of attributes.
- Updated `DeltaMergeDSV2Job` and `DeltaDSV2Spec` accordingly
- Fixed a minor bug in the `DeltaMergeDSV2Job` (missing string interpolation)

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
@@ -18,6 +18,7 @@ package za.co.absa.spline.harvester.builder
 
 import org.apache.spark.sql.catalyst.{expressions => sparkExprssions}
 import za.co.absa.spline.harvester.IdGeneratorsBundle
+import za.co.absa.spline.harvester.builder.OperationNodeBuilder.Attributes
 import za.co.absa.spline.harvester.builder.plan.PlanOperationNodeBuilder.OperationId
 import za.co.absa.spline.producer.model.{Attribute, FunctionalExpression, Literal}
 
@@ -33,10 +34,10 @@ trait OperationNodeBuilder {
   def addChild(childBuilder: OperationNodeBuilder): Unit = childBuilders :+= childBuilder
   protected def resolveAttributeChild(attribute: sparkExprssions.Attribute): Option[sparkExprssions.Expression] = None
 
-  protected def inputAttributes: Seq[Seq[Attribute]] = childBuilders.map(_.outputAttributes)
+  protected def inputAttributes: Seq[Attributes] = childBuilders.map(_.outputAttributes)
   protected def idGenerators: IdGeneratorsBundle
 
-  def outputAttributes: Seq[Attribute]
+  def outputAttributes: Attributes
 
   def childIds: Seq[OperationId] = childBuilders.map(_.operationId)
 
@@ -45,4 +46,8 @@ trait OperationNodeBuilder {
   def literals: Seq[Literal]
 
   def outputExprToAttMap: Map[sparkExprssions.ExprId, Attribute]
+}
+
+object OperationNodeBuilder {
+  type Attributes = Seq[Attribute]
 }
--- a/core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
@@ -18,7 +18,7 @@ package za.co.absa.spline.harvester.builder
 
 import org.apache.spark.sql.catalyst.{expressions => sparkExprssions}
 import za.co.absa.spline.harvester.IdGeneratorsBundle
-import za.co.absa.spline.harvester.builder.OperationNodeBuilder.Attributes
+import za.co.absa.spline.harvester.builder.OperationNodeBuilder.IOAttributes
 import za.co.absa.spline.harvester.builder.plan.PlanOperationNodeBuilder.OperationId
 import za.co.absa.spline.producer.model.{Attribute, FunctionalExpression, Literal}
 
@@ -34,10 +34,10 @@ trait OperationNodeBuilder {
   def addChild(childBuilder: OperationNodeBuilder): Unit = childBuilders :+= childBuilder
   protected def resolveAttributeChild(attribute: sparkExprssions.Attribute): Option[sparkExprssions.Expression] = None
 
-  protected def inputAttributes: Seq[Attributes] = childBuilders.map(_.outputAttributes)
+  protected def inputAttributes: Seq[IOAttributes] = childBuilders.map(_.outputAttributes)
   protected def idGenerators: IdGeneratorsBundle
 
-  def outputAttributes: Attributes
+  def outputAttributes: IOAttributes
 
   def childIds: Seq[OperationId] = childBuilders.map(_.operationId)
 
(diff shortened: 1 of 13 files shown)


core/src/main/scala/za/co/absa/spline/harvester/builder/plan/MergeIntoNodeBuilder.scala
```diff
@@ -29,7 +29,14 @@ class MergeIntoNodeBuilder
...
 
-  private lazy val mergeInputs: Seq[Seq[Attribute]] = inputAttributes.transpose
+  private lazy val mergeInputs: Seq[Seq[Attribute]] = {
+    val Seq(srcAttrs, trgAttrs) = inputAttributes
+    val srcAttrsByName = srcAttrs.map(a => a.name -> a).toMap
+    trgAttrs.map(trg => {
+      val src = srcAttrsByName.get(trg.name)
+      Seq(trg) ++ src
+    })
+  }
```
cerveada: What happens when the names don't match:
```
  | WHEN MATCHED THEN
  |   UPDATE SET
  |     name = src.full_name
```

I would also rename `src` -> `maybeSrc`

wajda: > I would also rename `src` -> `maybeSrc`

renamed

> What happens when the names don't match

Well, I guess nothing. Since there is no reference from a target attribute to the expression in Spark exec plan (or at least I can't see it) we cannot do much about it. For example from the below screenshot how would I infer dependencies between `full_name -> name` or `code -> name`? The agent could guess based on e.g. Levenshtein distance, but that would be another feature.

image

Actually, now thinking about it more I realized that even if names match it doesn't mean there is a dependency as there could be mutual renaming in the query (`UPDATE SET dst.a = src.b, dst.b = src.a`).

Do you have any better suggestion how to do it properly?

cerveada: > Actually, now thinking about it more I realized that even if names match it doesn't mean there is a dependency as there could be mutual renaming in the query (UPDATE SET dst.a = src.b, dst.b = src.a).

Exactly. I think the only way to do this properly is to get the attribute pairing from UPDATE SET condition.

wajda: I've re-implemented the dependency resolver algorithm, it should now cover all the cases discussed above, plus sub-attribute references. Please check.

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
```diff
@@ -46,3 +47,7 @@ trait OperationNodeBuilder {
 
   def outputExprToAttMap: Map[sparkExprssions.ExprId, Attribute]
 }
+
+object OperationNodeBuilder {
+  type Attributes = Seq[Attribute]
```
cerveada: I feel like this will only decrease the readability. I don't see any problems with using `Seq[Attribute]` as a type.

core/src/main/scala/za/co/absa/spline/harvester/builder/plan/MergeIntoNodeBuilder.scala
```diff
@@ -16,56 +16,94 @@
...
-    val attr1 = attributes.head
-    val idRefs = attributes.map(a => AttrRef(a.id))
-    Attribute(
-      id = idGenerators.attributeIdGenerator.nextId(),
-      dataType = attr1.dataType,
-      childRefs = idRefs,
-      extra = Map(CommonExtras.Synthetic -> true),
-      name = attr1.name
-    )
+  override lazy val outputAttributes: Seq[Attribute] = {
```
cerveada: Will this method work when the input attributes are synthetic? I think in this case the proper way is to take `inputAttributes` and use their id's because they may differ from the Spark ones.

core/src/main/scala/za/co/absa/spline/harvester/builder/OperationNodeBuilder.scala
wajda: It was primarily introduced to get rid of `Seq[Seq[Attribute]] `in other places.

cerveada (approved):

PR #617, opened 2023-02-24
issue #616 YAML config to convert single-valued non-string properties…

### LINKED ISSUE #616: YAML config breaks on non-string values
This is a continuation of #610 

> This is fixed in commons-configuration:1.7 https://issues.apache.org/jira/browse/CONFIGURATION-432
> 
> Unfortunately, Spark 2.4 depends on version 1.6
> 
> There already were other issues with YAML as version 1.2 being interpreted as float and there may be more.
Other properties sources just treat everything as string and do the conversion once it is asked for. Maybe we could use Yaml the same way.

wajda:
… to String for better interoperability with other types of `Configuration`.

fixes #616

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/conf/YAMLConfiguration.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/conf/YAMLConfiguration.scala
@@ -58,8 +58,10 @@ class YAMLConfiguration(yaml: String)
         case null => // ignore entries with `null` value
         case m: ju.Map[_, _] =>
           populateRecursively(dstKey, m.asInstanceOf[ju.Map[String, AnyRef]], dstMap)
+        case xs: ju.List[_] =>
+          dstMap.put(dstKey, xs)
         case v =>
-          dstMap.put(dstKey, v)
+          dstMap.put(dstKey, v.toString)
       }
     })
     dstMap
--- a/core/src/test/scala/za/co/absa/spline/harvester/conf/YAMLConfigurationTest.scala
+++ b/core/src/test/scala/za/co/absa/spline/harvester/conf/YAMLConfigurationTest.scala
@@ -34,6 +34,22 @@ class YAMLConfigurationTest extends ReadOnlyConfigurationTest {
       |    - c
       |  ~: null_key
       |  null_value:
+      |
+      |  # ------------------------------
+      |  # for value type sensitive tests
+      |  # ------------------------------
+      |
+      |  decimalAsNumber: 1.2
+      |  decimalAsString: "1.2"
+      |  integerAsNumber: 42
+      |  integerAsString: "42"
+      |  booleanAsBoolean: true
+      |  booleanAsString: "true"
+      |  heterogeneousValues:
+      |    - 1.2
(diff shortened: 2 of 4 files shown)


core/src/test/scala/za/co/absa/spline/harvester/conf/YAMLConfigurationTest.scala
```diff
@@ -47,4 +63,31 @@ class YAMLConfigurationTest extends ReadOnlyConfigurationTest {
...
+
+  test("get string value as String") {
+    givenConf getString "spline.x.y" shouldEqual "foo"
+    givenConf getString "spline.w.z" shouldEqual "bar"
+  }
+
+  test("get string value as another type") {
+    givenConf getDouble "spline.decimalAsString" shouldEqual 1.2d
+    givenConf getInt "spline.integerAsString" shouldEqual 42
+    givenConf getBoolean "spline.booleanAsString" shouldEqual true
```
cerveada: I think `type` to `type` should be also tested

like:
```
givenConf getDouble "spline.integerAsNumber" shouldEqual 1.2d
```

core/src/main/scala/za/co/absa/spline/harvester/conf/YAMLConfiguration.scala
```diff
@@ -58,8 +58,10 @@ class YAMLConfiguration(yaml: String)
         case null => // ignore entries with `null` value
         case m: ju.Map[_, _] =>
           populateRecursively(dstKey, m.asInstanceOf[ju.Map[String, AnyRef]], dstMap)
+        case xs: ju.List[_] =>
+          dstMap.put(dstKey, xs)
```
cerveada, line 62: Shouldn't the list elements be also converted to String?

core/src/test/scala/za/co/absa/spline/harvester/conf/YAMLConfigurationTest.scala
wajda: It was too obvious for me to test, but interestingly enough I spotted an error in this one :open_mouth:

core/src/main/scala/za/co/absa/spline/harvester/conf/YAMLConfiguration.scala
wajda: The main reason why I decided not to convert the list items to string was the signature of the `Configuration.getList()` method, that returns a list of objects (not strings). Having an alternative method `getStringArray()` that returns array of strings I assumed that `getList` could be used to return a collection of typed values (if the underlying implementation supports one).

core/src/test/scala/za/co/absa/spline/harvester/conf/YAMLConfigurationTest.scala
wajda: nah... that was a glitch :laughing:

wajda: added a unit test

cerveada (approved):
LGTM

PR #618, opened 2023-02-24
Adding feature to include token generation for long process of spark.

rupesh3020:
As we have feature to add HTTP headers in the header configuration but this token expires in one hour and spline does not have a mechanism to recreate token and make requests.

This feature adds the functionality by adding a new configuration authentication, where we can specify 
* clientid, clientsecret and other details to get tokens on the fly

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
@@ -28,7 +28,7 @@ trait Authentication {
   def createRequest(httpRequest: HttpRequest,authentication: scala.collection.immutable.Map[String,String]):HttpRequest
 }
 
-case class NoAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication with Logging {
+case class NoAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication{
   override def createRequest(httpRequest: HttpRequest, authentication: scala.collection.immutable.Map[String, String]): HttpRequest = httpRequest
 }
 case class ClientCredentialsAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication with Logging {
@@ -64,11 +64,10 @@ case class ClientCredentialsAuthentication(authentication: scala.collection.immu
         case Some(map: scala.collection.immutable.Map[String, Any]) =>
           val token = map.getOrElse("access_token", "").toString
           if (token.nonEmpty) {
-            val expirationTime = Instant.now().plus(Duration.ofSeconds(map.getOrElse("expires_in", "0").toString.toLong))
+            val expirationTime = Instant.now().plus(Duration.ofSeconds(map.getOrElse("expires_in", 0).toString.toDouble.toInt))
             val newToken = Token(token,expirationTime)
             tokenCache.put("token",newToken)
           }
-          logInfo(token)
           token
         case _ =>
           throw new RuntimeException(failureMessage)
@@ -92,9 +91,8 @@ case class ClientCredentialsAuthentication(authentication: scala.collection.immu
   }
 }
 
-object AuthenticationFactory extends Logging {
+object AuthenticationFactory{
   def createAuthentication(authentication: scala.collection.immutable.Map[String, String]): Authentication = {
-    logInfo(authentication("grantType"))
     authentication("grantType") match {
       case "client_credentials" => ClientCredentialsAuthentication(authentication)
       case _ => NoAuthentication(authentication)
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
@@ -10,13 +10,16 @@ trait Authentication {
   def createRequest(httpRequest: HttpRequest,authentication: Map[String,String]):HttpRequest
 }
 
+case class NoAuthentication(authentication: Map[String, String]) extends Authentication {
+  override def createRequest(httpRequest: HttpRequest, authentication: Map[String, String]): HttpRequest = httpRequest
+}
 case class ClientCredentialsAuthentication(authentication: Map[String, String]) extends Authentication {
   private val tokenCache: mutable.Map[String, (String, Instant)] = mutable.Map.empty[String, (String, Instant)]
-  private val clientId: String = ""
-  private val clientSecret: String = ""
-  private val tokenUrl: String = ""
-  private val grantType: String = ""
-  private val scope: String = ""
+  private var clientId: String = ""
+  private var clientSecret: String = ""
+  private var tokenUrl: String = ""
+  private var grantType: String = ""
+  private var scope: String = ""
 
   private def isTokenExpired(expirationTime: Instant): Boolean = {
     Instant.now().isAfter(expirationTime.minusSeconds(300) )
@@ -56,10 +59,11 @@ case class ClientCredentialsAuthentication(authentication: Map[String, String]) 
     authentication.get("mode") match {
       case Some("enabled") => {
         if (authentication.contains("clientId") && authentication.contains("clientSecret") && authentication.contains("tokenUrl") && authentication.contains("scope")) {
-          val clientId = authentication("clientId")
-          val clientSecret = authentication("clientSecret")
-          val tokenUrl = authentication("tokenUrl")
-          val scope = authentication("scope")
+          clientId = authentication("clientId")
+          clientSecret = authentication("clientSecret")
+          tokenUrl = authentication("tokenUrl")
+          scope = authentication("scope")
+          grantType = authentication("grantType")
           val token = getToken(clientId, clientSecret, tokenUrl, scope)
           httpRequest.header("Authorization", s"Bearer $token")
         } else {
@@ -75,7 +79,7 @@ object AuthenticationFactory {
   def createAuthentication(authentication: Map[String, String]): Authentication = {
     authentication("grantType") match {
       case "client_credentials" => ClientCredentialsAuthentication(authentication)
-      case _ => throw new IllegalArgumentException("Invalid grant type.")
+      case _ => NoAuthentication(authentication)
     }
   }
 }
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
@@ -14,7 +14,8 @@ case class NoAuthentication(authentication: Map[String, String]) extends Authent
   override def createRequest(httpRequest: HttpRequest, authentication: Map[String, String]): HttpRequest = httpRequest
 }
 case class ClientCredentialsAuthentication(authentication: Map[String, String]) extends Authentication {
-  private val tokenCache: mutable.Map[String, (String, Instant)] = mutable.Map.empty[String, (String, Instant)]
+  case class Token(tokenValue: String, expirationTime: Instant)
+  private var tokenCache: Map[String, Token] = Map.empty[String, Token]
   private var clientId: String = ""
   private var clientSecret: String = ""
   private var tokenUrl: String = ""
@@ -27,9 +28,9 @@ case class ClientCredentialsAuthentication(authentication: Map[String, String]) 
 
   private def getToken(clientId: String, clientSecret: String, tokenUrl: String,scope: String): String = {
     val cachedToken= tokenCache.get("token")
-    if (cachedToken.isDefined && !isTokenExpired(cachedToken.get._2)) {
+    if (cachedToken.isDefined && !isTokenExpired(cachedToken.get.expirationTime)) {
       // Token found in cache and not expired, return it
-      cachedToken.get._1
+      cachedToken.get.tokenValue
     } else {
       val resp = scalaj.http.Http(tokenUrl)
         .postForm(Seq(
--- a/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
@@ -1,0 +1,82 @@
+package za.co.absa.spline.harvester.dispatcher.httpdispatcher.rest
+
+import java.time.{Duration, Instant}
+
+import scalaj.http.{Base64, HttpRequest}
+
+import scala.collection.mutable
+
+trait Authentication {
+  def createRequest(httpRequest: HttpRequest,authentication: Map[String,String]):HttpRequest
+}
+
+case class ClientCredentialsAuthentication(authentication: Map[String, String]) extends Authentication {
+  private val tokenCache: mutable.Map[String, (String, Instant)] = mutable.Map.empty[String, (String, Instant)]
+  private val clientId: String = ""
+  private val clientSecret: String = ""
+  private val tokenUrl: String = ""
+  private val grantType: String = ""
+  private val scope: String = ""
+
+  private def isTokenExpired(expirationTime: Instant): Boolean = {
+    Instant.now().isAfter(expirationTime.minusSeconds(300) )
+  }
+
+  private def getToken(clientId: String, clientSecret: String, tokenUrl: String,scope: String): String = {
+    val cachedToken= tokenCache.get("token")
+    if (cachedToken.isDefined && !isTokenExpired(cachedToken.get._2)) {
+      // Token found in cache and not expired, return it
+      cachedToken.get._1
+    } else {
+      val resp = scalaj.http.Http(tokenUrl)
+        .postForm(Seq(
+          "grant_type" -> "client_credentials",
+          "client_id" -> clientId,
+          "client_secret" -> clientSecret,
+          "scope" -> scope))
+        .asString
+
+      val jsonResp = scala.util.parsing.json.JSON.parseFull(resp.body)
+      jsonResp match {
+        case Some(map: Map[String, Any]) =>
+          val token = map.getOrElse("access_token", "").toString
+          if (token.nonEmpty) {
+            val expirationTime = Instant.now().plus(Duration.ofSeconds(map.getOrElse("expires_in", "0").toString.toLong))
+            tokenCache.put("token", (token, expirationTime))
+          }
+          token
+        case _ =>
+          throw new RuntimeException("Failed to retrieve token from response")
+
+      }
+    }
+  }
+
+  override def createRequest(httpRequest: HttpRequest, authentication: Map[String, String]): HttpRequest ={
+    authentication.get("mode") match {
+      case Some("enabled") => {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
```diff
@@ -26,14 +26,41 @@ import java.io.ByteArrayOutputStream
 import java.util.zip.GZIPOutputStream
 import javax.ws.rs.HttpMethod
 
-class RestEndpoint(val request: HttpRequest) {
+class RestEndpoint(val request: HttpRequest,val authentication: Option[Map[String, String]]) {
 
-  def head(): HttpResponse[String] = request
+  def withAuth(): HttpRequest = {
+    if(authentication.contains("clientId") && authentication.contains("clientSecret") && authentication.contains("tokenUrl") && authentication.contains("scope")) {
```
cerveada: Those `contains` will always fail, they are call on the option not on the map.

I think the wrapping Option is not needed here, we can just use empty map when no authentication is required.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
```diff
@@ -26,14 +26,41 @@ import java.io.ByteArrayOutputStream
...
+      .postForm(Seq(
+        "grant_type" -> "client_credentials",
+        "client_id" -> clientId,
+        "client_secret" -> clientSecret))
+      .asString
+
+    val jsonResp = scala.util.parsing.json.JSON.parseFull(resp.body)
+    jsonResp match {
+      case Some(map: Map[String, Any]) => map.getOrElse("access_token", "").toString
+      case _ => ""
```
cerveada: Why not fail here? Without the token, the request will fail anyway, right?

cerveada:
Is this implementing Bearer authentification? (https://datatracker.ietf.org/doc/html/rfc6750)

I think it would make sense to have `authentification.type` property to allow other types in the future.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
```diff
@@ -26,14 +26,41 @@ import java.io.ByteArrayOutputStream
...
+    }
+  }
+      
+  def head(): HttpResponse[String] = withAuth()
     .method(HttpMethod.HEAD)
     .asString
 
   def post(data: String, contentType: String, enableRequestCompression: Boolean): HttpResponse[String] = {
-    val jsonRequest = request
+    val jsonRequest = withAuth()
```
cerveada: This will request a new token before each request. Isn't it enough to get the token just once?

cerveada:
Currently, the build is failing because test do not compile. The tests must be modified to accommodate the changes in the dispatcher.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
wajda: @cerveada by "once" you mean at the first method call and cache it? The point here is that we cannot obtain it on the init time because it can expire by the time the job has finished.

cerveada: Well, then it depends on how quickly it expires. Maybe it is not wort it to do the caching.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
rupesh3020: Sure, I have changed from option to a map

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
rupesh3020: As we know that the request is only made when there is an action. This creates token when post is done. We can also put a function to check the expiry. Let me know if it is ok.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
rupesh3020: Done , I have raised a exception if token is not matched

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
rupesh3020: Done, now it checks if token is expired or not. If it is expired then it creates token again

rupesh3020:
> Is this implementing Bearer authentification? (https://datatracker.ietf.org/doc/html/rfc6750)
> 
> I think it would make sense to have `authentification.type` property to allow other types in the future.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestEndpoint.scala
rupesh3020: @wajda @cerveada I have tested it. please let me know if something else is required here

rupesh3020:
> Currently, the build is failing because test do not compile. The tests must be modified to accommodate the changes in the dispatcher.

@cerveada Tests are fixed and i have tested the jar, it works fine. can you please check

cerveada:
@rupesh3020 thanks for fixing the tests and other issues, I will take a look.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/RestClient.scala
```diff
@@ -41,14 +41,16 @@ object RestClient extends Logging {
...
+    headers: Map[String, String],
+    authentication: Map[String, String]
   ): RestClient = {
 
     logDebug(s"baseURL = $baseURL")
     logDebug(s"connTimeout = $connTimeout")
     logDebug(s"readTimeout = $readTimeout")
     logDebug(s"disableSslValidation = $disableSslValidation")
     logDebug(s"headers = $headers")
+    logInfo(s"authe= $authentication")
```
cerveada: We should be careful to not log any secrets

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+      case _ => httpRequest
+    }
+  }
+}
+
+object AuthenticationFactory{
+  def createAuthentication(authentication: scala.collection.immutable.Map[String, String]): Authentication = {
+    authentication("grantType") match {
```
cerveada: This will throw an exception when the map is empty. I would check for empty map and if it is empty I would use NoAuthentication

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,102 @@
...
+  override def createRequest(httpRequest: HttpRequest, authentication: scala.collection.immutable.Map[String, String]): HttpRequest = httpRequest
+}
+case class ClientCredentialsAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication with Logging {
+  case class Token(tokenValue: String, expirationTime: Instant)
+  private val tokenCache: scala.collection.mutable.Map[String, Token] = scala.collection.mutable.Map.empty[String, Token]
+  private val clientId: String = authentication.getOrElse("clientId","")
+  private val clientSecret: String = authentication.getOrElse("clientSecret","")
+  private val tokenUrl: String = authentication.getOrElse("tokenUrl","")
+  private val grantType: String = authentication.getOrElse("grantType","")
+  private val scope: String = authentication.getOrElse("scope","")
```
cerveada: For mandatory properties, use just ```authentication("clientId")``` it will fail when the property is missing, but that's what we want if it's mandatory.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
cerveada: for `mode` if it's disabled, just use NoAuthentication

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,102 @@
...
+          throw new RuntimeException(failureMessage)
+
+      }
+    }
+  }
+
+  override def createRequest(httpRequest: HttpRequest, authentication: scala.collection.immutable.Map[String, String]): HttpRequest ={
+    authentication.get("mode") match {
+      case Some("enabled") => {
+        if (authentication.contains("clientId") && authentication.contains("clientSecret") && authentication.contains("tokenUrl") && authentication.contains("scope")) {
```
cerveada: All those checks should be done already (when the class is instantiated) here it should be just get the token and put it in the header.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,102 @@
...
+  private val tokenCache: scala.collection.mutable.Map[String, Token] = scala.collection.mutable.Map.empty[String, Token]
+  private val clientId: String = authentication.getOrElse("clientId","")
+  private val clientSecret: String = authentication.getOrElse("clientSecret","")
+  private val tokenUrl: String = authentication.getOrElse("tokenUrl","")
+  private val grantType: String = authentication.getOrElse("grantType","")
+  private val scope: String = authentication.getOrElse("scope","")
+
+  private def isTokenExpired(expirationTime: Instant): Boolean = {
+    Instant.now().isAfter(expirationTime.minusSeconds(300) )
+  }
```
cerveada: Good idea to extract the expiration checking logic into a separated method. I would go even further and use a `isTokenValid` method where you can check both, that it is defined and the expiration.

rupesh3020:
@cerveada Issues fixed. Can you please check once. I have done functional testing

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,105 @@
...
+  def createRequest(httpRequest: HttpRequest,authentication: scala.collection.immutable.Map[String,String]):HttpRequest
+}
+
+case class NoAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication{
+  override def createRequest(httpRequest: HttpRequest, authentication: scala.collection.immutable.Map[String, String]): HttpRequest = httpRequest
+}
+
+case class ClientCredentialsAuthentication(authentication: scala.collection.immutable.Map[String, String]) extends Authentication with Logging {
+  case class Token(tokenValue: String, expirationTime: Instant)
+  private val tokenCache: scala.collection.mutable.Map[String, Token] = scala.collection.mutable.Map.empty[String, Token]
```
cerveada: Since you have just one token, no need to use a Map.

Please use `private var tokenCache: Option[Token] = None`

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,105 @@
...
+  private val clientSecret: String = authentication("clientSecret")
+  private val tokenUrl: String = authentication("tokenUrl")
+  private val grantType: String = authentication("grantType")
+  private val scope: String = authentication("scope")
+  val failureMessage = "Failed to retrieve token from response"
+
+  private def isTokenInvalid: Boolean = {
+    val cachedToken = tokenCache.get("token")
+    cachedToken.isEmpty || Instant.now().isAfter(cachedToken.getOrElse(throw new IllegalArgumentException(failureMessage)).expirationTime.minusSeconds(300))
+  }
```
cerveada: Since you already checked `cachedToken.isEmpty` you can do `cachedToken.get` instead of `getOrElse`. there is no need for exception here. Token not being set simply means it is invalid.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,105 @@
...
+
+object AuthenticationFactory extends Logging {
+  def createAuthentication(authentication: scala.collection.immutable.Map[String, String]): Authentication = {
+    if (authentication.isDefinedAt("mode") && authentication.isDefinedAt("grantType")) {
+      authentication("mode") match {
+          case "enabled" => authentication("grantType") match {
+                                  case "client_credentials" => ClientCredentialsAuthentication(authentication)
+                                  case _ =>
+                                    logInfo(s"$authentication('grantType') not implemented")
+                                    NoAuthentication(authentication)
```
cerveada: too much indentation

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,105 @@
...
+          "client_secret" -> clientSecret,
+          "scope" -> scope))
+        .asString
+
+      val jsonResp = scala.util.parsing.json.JSON.parseFull(resp.body)
+      jsonResp match {
+        case Some(map: scala.collection.immutable.Map[String, Any]) =>
+          val token = map.getOrElse("access_token", "").toString
+          if (token.nonEmpty) {
+            val expirationTime = Instant.now().plus(Duration.ofSeconds(map.getOrElse("expires_in", 0).toString.toDouble.toInt))
```
cerveada: `toString.toDouble.toInt`
why are all those conversions needed?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
cerveada: When `authentication.mode` is `enabled` and some mandatory parameter is missing, we should throw an exception.

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
rupesh3020: It fails in static code analysis. Both in using var and using get . Can you suggest ?

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
rupesh3020: var is rejected by code analyser

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
cerveada: That is OK, generally using var and get on Option should be avoided, but in this case the cache must be mutable. And calling get on Option is OK if you know that it contains value.

rupesh3020: @cerveada done.. thanks for comment

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,104 @@
...
+      val jsonResp = scala.util.parsing.json.JSON.parseFull(resp.body)
+      jsonResp match {
+        case Some(map: scala.collection.immutable.Map[String, Any]) =>
+          val token = map.getOrElse("access_token", "").toString
+          if (token.nonEmpty) {
+            val expirationTime = Instant.now().plus(Duration.ofSeconds(map.getOrElse("expires_in", 0).toString.toLong))
+            val newToken = Token(token, expirationTime)
+            tokenCache = Some(newToken)
+          }
+          token
```
cerveada: In this case `access_token` is mandatory in the response(https://www.rfc-editor.org/rfc/rfc6749#section-5.1), so you may just assume it is there.

No need for `map.getOrElse("access_token", "")` or checking `token.nonEmpty`.

Just use `map("access_token")`

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
```diff
@@ -0,0 +1,104 @@
...
+        case _ =>
+          throw new RuntimeException(failureMessage)
+      }
+    } else {
+      tokenCache.getOrElse(throw new RuntimeException(failureMessage)).tokenValue
```
cerveada: in this place `isTokenInvalid` was already checked, so you can safely do `tokenCache.get.tokenValue`

rupesh3020: @cerveada done

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
rupesh3020: Done

cerveada (approved):
LGTM

core/src/main/scala/za/co/absa/spline/harvester/dispatcher/httpdispatcher/rest/AuthenticationFactory.scala
rupesh3020: @cerveada  Can you please merge if everything looks good as we are waiting for this feature.

wajda:
Sorry guys, I was really buried with other stuff recently, but I really would like to have a quick look it it. Give me an hour or two, I'll check it out. In general, yes, that's awesome feature, I'm not going to delay merge.

wajda:
@rupesh3020 Thank you very much for your contribution!. We've made a few changes to your code and merged it as another pull request - #631. So I'm closing this one for that reason.

PR #631, opened 2023-03-24
pr/618

wajda:
This is a formatted and slightly refactored version of the PR #618 by @rupesh3020

Excepts for some minor code formatting the changes I made are concentrated in the `AuthenticationFactory.scala` file:
1. Renamed `AuthenticationFactory.scala` to `authentication.scala` according to the Scala style guide for files that contain more classes.
2. Replaced deprecated Scala JSON parser with the agent internal `HarversterJsonSerDe` and simplified the connected code.
3. Replaced unnecessary case classes with objects and normal classes.
4. Encapsulated all interactions with the mutable variable `tokenCache` in one method for clarity.
5. Made the rest of the code more concise and self-describing.


core/src/main/resources/spline.default.yaml
```diff
@@ -68,6 +68,16 @@ spline:
             # Should the payload be compressed.
             # If unspecified, the request compression is negotiated with the endpoint, otherwise it's disabled.
             requestCompression: # true | false
+
+            # Authentication config
+            authentication:
+                mode: enabled
```
cerveada: Shouldn't the default be disabled?

cerveada (approved):
LGTM

PR #670, opened 2023-04-27
#669 - Add code coverage support improve

### LINKED ISSUE #669: Improve code-coverage & add GH check action
## Background
Improve or add code coverage support to be able measure code quality.

## Feature
New ability to measure current code coverage as one of QA metrics.
Add GH action to check changed file coverage.

miroslavpojer:
- Used Absa fork of JaCoCo solution - with scala method filtering.
- Added support for code coverage GitHub action check as part of PR.

Closes #669

### DIFF
--- a/.github/workflows/ci-check-jacoco.yml
+++ b/.github/workflows/ci-check-jacoco.yml
@@ -1,0 +1,39 @@
+name: CI check JaCoCo code-coverage
+
+on:
+  pull_request:
+    branches: [ develop ]
+    types: [ opened, edited, synchronize, reopened ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: "adopt@1.8"
+      - name: Build and run tests
+        run: mvn verify -Dcode-coverage
+      - name: Add coverage to PR
+        id: jacoco
+        uses: madrapps/jacoco-report@v1.3
+        with:
+          paths: ${{ github.workspace }}/core/target/site/jacoco/jacoco.xml
+          token: ${{ secrets.GITHUB_TOKEN }}
+          min-coverage-overall: 0.0
+          min-coverage-changed-files: 80.0
+          title: JaCoCo code coverage report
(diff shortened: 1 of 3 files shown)


cerveada:
Does this coverage take into account the integration-tests or just the standard unit tests?

miroslavpojer:
> Does this coverage take into account the integration-tests or just the standard unit tests?

Total project coverage is measured and counted only for core module in project. For coverage measuring are used tests which are inside of this module.

cerveada:
This project is somewhat atypical in that it relies heavily on integration tests. A huge portion of logic here is matching and extracting objects coming from Spark, and we need to test that it works for every version of Spark and Scala supported. 

So unit test are useless for a big part of the code, we used them to test the internal logic of course, but that is the smaller part of the codebase.

I think we need to either include the integration test in the metric (not sure if it's possible) or drop the requirement for minimal coverage percentage.

miroslavpojer:
> This project is somewhat atypical in that it relies heavily on integration tests. A huge portion of logic here is matching and extracting objects coming from Spark, and we need to test that it works for every version of Spark and Scala supported.
> 
> So unit test are useless for a big part of the code, we used them to test the internal logic of course, but that is the smaller part of the codebase.
> 
> I think we need to either include the integration test in the metric (not sure if it's possible) or drop the requirement for minimal coverage percentage.

I understand. You are missing crossbuild feature in maven.
I do not know the way how to connect two modules to measure code coverage this way. The basic idea is to measure one module only.
I see three options:
- switch project to sbt where is cross build present and applied by several projects in AbsaOSS
- drop entire jacoco GH action
- lower GH blocking percentage to acceptable value

cerveada:
Cross-build is not a big factor here. The core issue is that we care about the test coverage (unit + integration) but we measure unit only.

miroslavpojer:
> Cross-build is not a big factor here. The core issue is that we care about the test coverage (unit + integration) but we measure unit only.

Yes, until integration test will be outside of measured module.

cerveada:
Spline team decision is to not use the GitHub action since unit test coverage (alone) is not a useful metric for this project, but let's keep the maven goal there for anyone if they need to check this.

cerveada (approved):

wajda (approved):

PR #694, opened 2023-06-08
Use shaded jackson artifacts

### LINKED ISSUE #675: Replace json4s by other json SerDe
Since json4s seems to be causing compatibility issues between different versions of Spark and shading, it seems to be difficult (see #667) it is worth considering to replace it by different library.

Ideal features of the library would be:
- Support for Scala 2.11, 2.12, 2.13
- Stable - Ideally a library that keeps each major release in a different package, [described here](https://github.com/rallyhealth/sbt-shading#chapter-3-a-solution-emerges)
- Automatic mapping of case classes to json

Some candidates:

https://github.com/spray/spray-json
no automapping

https://github.com/rallyhealth/weePickle
no automapping

cerveada:
Replaces Json4s by Jackson

Tested by running all examples and comparing the results. Differences:
- `_typeHint` is last object param instead of first
- params map now may contain key value pairs with value=null

Both doesn't seem to be a big issue, other than that I haven't found any differences in resulting jsons.

fixes #675


integration-tests/pom.xml
```diff
@@ -123,7 +123,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>2.15.0</version>
+            <version>2.15.2</version>
         </dependency>
```
wajda: I guess we can delete the `jackson-annotations` dependency entirely from here.

cerveada: That's what I thought as well, but the test containers need it to run.

wajda (approved):

PR #715, opened 2023-06-28
Bugfix/spline spark agent 714 classpath collision

### LINKED ISSUE #714: Classpath collision on ABSA Commons
```
23/06/28 12:39:27 INFO ApplicationMaster: Unregistering ApplicationMaster with FAILED
    (diag message: User class threw exception: java.lang.NoSuchMethodError: 
    za.co.absa.commons.config.ConfigurationImplicits$.ConfigurationMapWrapper(Lorg/apache/commons/configuration/Configuration;)Lorg/apache/commons/configuration/Configuration;
```

wajda:
fixes #714

Relocated:
  - `za.co.absa.commons:commons`
  - `io.github.classgraph:glassgraph`
  - `com.fasterxml.uuid:java-uuid-generator`
  - `org.scalaj:scalaj-http`
  - `com.lihaoyi:fastparse`

Moved from the _root_ module to the _integration-tests_ module:
  - `com.github.bigwheel:util-backports`

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -94,6 +94,11 @@
 
         <dependency>
             <groupId>za.co.absa.shaded</groupId>
+            <artifactId>absa-shaded-jackson</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>za.co.absa.shaded</groupId>
             <artifactId>absa-shaded-jackson-module-scala_${scala.binary.version}</artifactId>
             <version>0.0.2</version>
         </dependency>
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -87,11 +92,6 @@
 
         <!-- dependencies shaded by us-->
 
-        <dependency>
-            <groupId>za.co.absa.shaded</groupId>
-            <artifactId>absa-shaded-jackson</artifactId>
-            <version>0.0.1</version>
-        </dependency>
```
cerveada: I felt like we should declare dependencies that we use directly, even though they are used transitively as well. Do you not agree?

cerveada:
Does this break tests when running from IDE?

cerveada (approved):

wajda:
> Does this break tests when running from IDE?

No, in root pom, there is only a shade plugin config, but it's only executed in bundles.
I added the comment for clarity.

PR #728, opened 2023-07-30
Feature/spline spark agent 726 multi arch docker

### LINKED ISSUE #726: Multi-arch docker builds
Update to root POM ver 1.0.11 to support multi-arch docker builds

wajda:
fixes #726 

- update root pom to version 1.0.11 to support multi-arch builds
- redo examples running. Get rid of Maven, Ant, simplify the whole process.
- move Dockerfile and related POM configuration to the examples module.


examples/src/main/scala/za/co/absa/spline/issue/GitHub738Job.scala
```diff
@@ -33,8 +33,8 @@ object GitHub738Job extends SparkApp("GitHub spline-738") with Logging {
   val primarySourcePaths = primarySourceKeys.map(key => createMockCSV(s"Primary-$key"))
   val secondarySourcePaths = secondarySourceKeys.map(key => createMockCSV(s"Secondary-$key"))
 
-  val nIters = 20
-  val nAppends = 10
+  val nIters: Int = if (args.length > 0) args(0).toInt else 4 // 20
+  val nAppends: Int = if (args.length > 1) args(1).toInt else 2 // 10
```
wajda:
```suggestion
  val nIters: Int = if (args.length > 0) args(0).toInt else 2 // 20
  val nAppends: Int = if (args.length > 1) args(1).toInt else 2 // 10
```

examples/Dockerfile
```diff
@@ -14,39 +14,40 @@
 # limitations under the License.
 #
 
-FROM maven:3.6.0-jdk-8-slim
+FROM eclipse-temurin:11-jre-jammy
```
cerveada, line 17: What is the reason for using eclipse-temurin instead of the maven image?

examples/src/main/scala/za/co/absa/spline/issue/GitHub738Job.scala
```diff
@@ -33,8 +33,8 @@ object GitHub738Job extends SparkApp("GitHub spline-738") with Logging {
   val primarySourcePaths = primarySourceKeys.map(key => createMockCSV(s"Primary-$key"))
```
cerveada, line 39: It would be nice to have a short comment on this job, it's good to have a ticket number, but still...

the comment could describe what is the purpose of this job and link to the ticket, since it's not in agent github, so it may be otherwise confusing.

examples/Dockerfile
wajda: We do not need maven anymore. I was looking for the smallest JDK11 image that is available for both arm64 and amd64 platforms

examples/src/main/scala/za/co/absa/spline/issue/GitHub738Job.scala
wajda: It's in Spline's repo, the old one - AbsaOSS/spline#738
Ok, I'll make a comment

wajda: done. please check

examples/src/main/scala/za/co/absa/spline/issue/GitHub718Job.scala
```diff
@@ -20,6 +20,11 @@ import org.apache.commons.io.FileUtils
 import za.co.absa.commons.io.TempFile
 import za.co.absa.spline.SparkApp
 
+/**
+ * THis is a test scenario for the lineage graph alignment on the UI
```
cerveada:
```suggestion
 * This is a test scenario for the lineage graph alignment on the UI
```

cerveada (approved):

PR #736, opened 2023-08-16
minor: Fix sonarcube warnings

wajda:
Fix SonarCloud warnings

1. Add missing comments for empty methods
2. Remove unused definitions and commented out code
3. Simplify boolean expressions
4. Mute false security alarms on leaked secrets (dummy passwords used in unit tests for testing password processing)
5. Mute code duplication warnings in tests

### DIFF
--- a/integration-tests/src/main/scala/za/co/absa/spline/test/LineageWalker.scala
+++ b/integration-tests/src/main/scala/za/co/absa/spline/test/LineageWalker.scala
@@ -38,14 +38,9 @@ class LineageWalker(
 
   private def dependsOnRec(ref: AttrOrExprRef, id: String): Boolean = ref match {
     case AttrRef(attrIfd) =>
-      if (attrIfd == id) true
-      else dependsOnRec(attrMap(attrIfd).childRefs, id)
+      attrIfd == id || dependsOnRec(attrMap(attrIfd).childRefs, id)
     case ExprRef(exprId) =>
-      if (exprId == id) true
-      else {
-        if (litMap.contains("exprId")) false
-        else dependsOnRec(funMap(exprId).childRefs, id)
-      }
+      exprId == id || !litMap.contains("exprId") && dependsOnRec(funMap(exprId).childRefs, id)
   }
 
 }
--- a/integration-tests/src/main/scala/za/co/absa/spline/test/LineageWalker.scala
+++ b/integration-tests/src/main/scala/za/co/absa/spline/test/LineageWalker.scala
@@ -18,6 +18,14 @@ package za.co.absa.spline.test
 
 import za.co.absa.spline.producer.model._
 
+/**
+ * A class to walk through the execution plan.
+ *
+ * @param allOpMap A map of all operations in the execution plan.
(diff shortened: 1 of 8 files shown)


integration-tests/src/main/scala/za/co/absa/spline/test/LineageWalker.scala
```diff
@@ -38,14 +38,9 @@ class LineageWalker(
...
     case AttrRef(attrIfd) =>
-      if (attrIfd == id) true
-      else dependsOnRec(attrMap(attrIfd).childRefs, id)
+      attrIfd == id || dependsOnRec(attrMap(attrIfd).childRefs, id)
     case ExprRef(exprId) =>
-      if (exprId == id) true
-      else {
-        if (litMap.contains("exprId")) false
-        else dependsOnRec(funMap(exprId).childRefs, id)
-      }
```
cerveada: It's shorter, but for me, the new version is harder to understand.

wajda: What about introducing an intermediate val? For example?
```scala
lazy val attrChildIds = attrMap(attrId).childRefs
lazy val anyChildRefsDepends = dependsOnRec(attrChildRefs, id)
attrIfd == id || anyChildRefsDepends
```

The main motivation was to comply with the Sonar [rule](https://sonarcloud.io/project/issues?resolved=false&severities=MINOR&id=AbsaOSS_spline-spark-agent&open=AX7fDAUrz4PBF_ZPoJ7s&tab=why).

Subjectively, I understand why this is usually recommended approach. First, it's shorter, but also it is more plural for the reader. There is less nesting and branching involved for the reader. You have entire resulted condition written in one sentence that IMO looks more natural, e.g. `condition = this or that and not something_else`. But I agree that a lot depends on naming. So I wouldn't hesitate choosing better names for functions, or introducing intermediate variables.

On a side note, I just noticed that `attrIfd` is probably a typo. Shouldn't it be `attrId`? Also `dependsOnRef` instead of `dependsOnRec`?

cerveada: For simple cases, It may be ok, but here the second condition is more complex. The nesting and branching is still there, it just now must happen inside the reader's head instead of in the code. The reader must consider short-circuiting, negate one expression and operator precedence, and then build a truth table to make sense of it. The Sonar just doesn't consider any of that.

Shorter is not better, I think we should aim to minimize cognitive load.

Why do you mean by `plural`?

`attrIfd` yes typo
`dependsOnRec` no it means recursive - just private name

wajda: plural -> fluent. Crazy typo :)

wajda: Let me refactor it a bit and I'll propose another variant.

cerveada (approved):
LGTM, but `dependsOn` is good candidate to unit test

PR #759, opened 2023-10-24
New `spline.pluginsEnabledByDefault` property

martinf-moodys:
This new property aims to fix a performance issue when scanning plugins via ClassGraph during Spline initialization.
This may take up to 30 seconds to call `ClassGraph.scan` from `AutoDiscoveryPluginRegistry`. By default, the property is set to true to keep the current behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,9 +698,12 @@ When one of these commands occurs spline will let you know by logging a warning.
 ### Plugin API
 
 Using a plugin API you can capture lineage from a 3rd party data source provider.
-Spline discover plugins automatically by scanning a classpath, so no special steps required to register and configure a plugin.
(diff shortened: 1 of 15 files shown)


martinf-moodys:
> This new property aims to fix a performance issue when scanning plugins via ClassGraph during Spline initialization. This may take up to 30 seconds to call `ClassGraph.scan` from `AutoDiscoveryPluginRegistry`. By default, the property is set to `true` to maintain the current behavior.

wajda:
I somehow missed this PR, sorry.
That's a good idea, but it contains a couple of things I'm not sure about. Let me review the PR thoroughly.

wajda:
This PR with slight changes is incorporated into #854

PR #739, opened 2023-08-30
spark agent #705 Spark 3.4 support

cerveada:
fixes #705

PR #793, opened 2024-03-08
Provide Spark 3.4 Support for Spline w/ Backwards Compatibility

### LINKED ISSUE #705: Spark 3.4 support
there seems to be binary incompatible changes in API of Delta and Spark SQL that Spline core compiled against Spark 2.4 version cannot work with. E.g. RDDPlugin

Todo:
1. Check how many places in code that actually incompatible and if there is a quick simple way to fix it.
2. Otherwise, we'll probably need to move toward Spark version specific builds (#604)

rycowhi:
fixes https://github.com/AbsaOSS/spline-spark-agent/issues/705

* Makes changes to failing tests to perform different behavior based on Spark version
* No new functionality actually required to meet compatibility 

TODO:

- [x] Create Agent Bundle POM 
- [x] Validate approach is sensible 
- [x] Rule out additional regression from changed Spark behavior
- [x] Request CI addition for Spark 3.4


core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/RDDPlugin.scala
```diff
@@ -39,14 +39,22 @@ class RDDPlugin(
...
+      val files = fsr.filePartitions.flatMap(_.files)
+      val uris = files.map(extractPath(_))
       ReadNodeInfo(SourceIdentifier(None, uris: _*), Map.empty)
     case hr: HadoopRDD[_, _] =>
       val partitions = ReflectionUtils.extractValue[Array[Partition]](hr, "partitions_")
       val uris = partitions.map(hadoopPartitionToUriString)
       ReadNodeInfo(SourceIdentifier(None, uris: _*), Map.empty)
   }
 
+  private def extractPath(file: PartitionedFile): String = {
```
rycowhi, line 51: These were some changes from the original PR #739 that this was based off of. I read mention of potential binary incompatibility yet these tests passed under both 3.3 + 3.4 profile. Any thoughts here?

integration-tests/pom.xml
```diff
@@ -267,6 +267,21 @@
...
+            </properties>
+            <dependencies>
+                <dependency>
+                    <groupId>org.apache.iceberg</groupId>
+                    <artifactId>iceberg-spark-runtime-3.4_${scala.binary.version}</artifactId>
+                    <version>1.3.1</version>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
+        </profile>
```
rycowhi, line 284: These changes were from the original PR this was based off of.

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
```diff
@@ -117,15 +122,41 @@ class BasicIntegrationTests extends AsyncFlatSpec
               .write.mode(Append).saveAsTable(tableName)
           }
 
-          (plan2, _) <- captor.lineageOf {
+          // Spark 3.4+ is creating 2 commands for both writes here so we need to ignore one
+          // We only want the one that is from CreateDataSourceTableAsSelectCommand
+          // The one we ignore here is an extra InsertIntoHadoopFsRelationCommand
+          // They can come out of order so we need to filter out which one is which.
+          (plan2, _) <- if (ver"$SPARK_VERSION" >= ver"3.4.0") {
```
rycowhi, line 129: I am not a fan of taking this approach but this does allow the tests to still pass - if we are comfortable with knowing that Spark is firing additional events here we get the same behavior.

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
```diff
@@ -117,15 +122,41 @@ class BasicIntegrationTests extends AsyncFlatSpec
...
+            captor.lineageOf {
+              Thread.sleep(5000)
+            }
+          } else Future[(ExecutionPlan, Seq[ExecutionEvent])](null, null)
         } yield {
           println("yield")
-          val writeUri = plan1.operations.write.outputSource
-          val readUri = plan2.operations.reads.head.inputSources.head
+
+          val writePlan = Seq(plan1, plan2)
```
rycowhi, line 149: There is some common code here on filtering out the correct options we want here between this and the Hive CTAS code - not sure how much level of DRY you all are looking to reduce here but this is potential.

integration-tests/src/test/scala/za/co/absa/spline/KafkaSinkSpec.scala
```diff
@@ -76,7 +76,7 @@ class KafkaSinkSpec
 
           (plan2, _) <- captor.lineageOf(
             reader
-              .option("subscribe", s"$topicName,anotherTopic")
+              .option("subscribe", s"$topicName")
```
rycowhi: So this one I found a bit odd - EmbeddedKafka was complaining about "anotherTopic" here not existing on this version specifically so I'm guessing Spark is doing something different here trying to read messages from that topic that it didn't before. 

This test is not looking for different behavior based on multiple topic names, but is this acceptable?

integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
```diff
@@ -500,7 +515,7 @@ class LineageHarvesterSpec extends AsyncFlatSpec
           plan should not be null
           event.durationNs should be(empty)
           event.error should not be empty
-          event.error.get.toString should include(s"path ${tmpLocal.toURI.toString.stripSuffix("/")} already exists")
+          event.error.get.toString.toLowerCase should include(s"path ${tmpLocal.toURI.toString.stripSuffix("/")} already exists")
```
rycowhi: It appears Spark changed the casing in their error here for some reason 😄

pom.xml
```diff
@@ -210,6 +214,7 @@
                 <groupId>com.github.cerveada</groupId>
                 <artifactId>scalatest-maven-plugin</artifactId>
                 <configuration>
+                    <stdout>I</stdout>
```
rycowhi, line 217: Only extra add from me here on this PR in this file - this option re-prints the failed tests w/ a tiny bit of log at the end of a run. Found it helpful while trying to figure out what needed to be fixed. Happy to remove if desired.

bundle-3.4/pom.xml
```diff
@@ -0,0 +1,1344 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rycowhi, line 1: Want to make sure I've done this right - 

1. Ran the jars-pommefizer on spark-3.4.1-bin-hadoop3
2. Copied over some basic pom setup info from bundle-3.3 (higher level project template
3. Moved dependencies from pommefizer output into dependencyManagement 
4. Filled in dependencies whose groupIds were blank from pommefizer (copied from 3.3, maybe a bug in pommefizer?)
5. Copied build plugins from 3.3
6. Manually removed versions from everything mentioned in dependencyManagement (this was a doozy 😄 )

integration-tests/src/test/scala/za/co/absa/spline/KafkaSinkSpec.scala
cerveada: I guess since this is not checked we can remove it, another option would be to add another write to `anotherTopic` as I understand it the problem is that it does not exist.

For now, though, I think removing `anotherTopic` is fine.

pom.xml
cerveada: If it changes the oputput only for failed tests, it can stay.

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/RDDPlugin.scala
cerveada: I think the changes from #739 are fine if the tests pass, no need to change it.

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
cerveada: So the Spark creates two plans where it used to create just one? What is the new root/write command that it creates? Spline should react only on write commands.

bundle-3.4/pom.xml
cerveada: Seems alright, @wajda what do you think?

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/RDDPlugin.scala
rycowhi: Sounds good!

pom.xml
rycowhi: Sounds good. From the ScalaTest docs:

```
I - show reminder of failed and canceled tests without stack traces
```

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
rycowhi: So this is what's confusing me a little bit - we get both a CreateDataSourceTableAsSelectCommand & InsertIntoHadoopFsRelationCommand (printed from LineageHarvester) in an a single Spark Action (and they aren't guaranteed to appear in the same order as well, hence why I needed to do that weird filter to find the right one).

I think I'm going to run a custom QueryExecutionListener and see if Spark itself happens to print out two actions as well. If not I'm going to be even more confused 😅

rycowhi: But yes - this didn't happen in < 3.4

integration-tests/src/test/scala/za/co/absa/spline/KafkaSinkSpec.scala
rycowhi: Good call this fixed it - I put the other topic back in, renamed the test a bit and added a sensible assertion. Thoughts?

integration-tests/src/test/scala/za/co/absa/spline/KafkaSinkSpec.scala
```diff
@@ -98,6 +109,7 @@ class KafkaSinkSpec
 
           plan2.operations.reads.head.extra("sourceType") shouldBe Some("kafka")
           plan2.operations.reads.head.inputSources should contain(s"kafka:$topicName")
+          plan2.operations.reads.head.inputSources should contain(s"kafka:$otherTopicName")
```
rycowhi, line 112: I think this is the intended behavior here @cerveada ?

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
rycowhi: Ok I think I've just confirmed that this is indeed Spark and not Spline doing this. I made a listener specifically for this test like so:

```scala
class TestListener extends QueryExecutionListener {

  override def onSuccess(funcName:  String, qe:  QueryExecution, durationNs:  Long): Unit = {
    println("A COMMAND JUST RAN")
    println(qe.commandExecuted.getClass.getCanonicalName)
  }
  override def onFailure(funcName:  _root_.scala.Predef.String, qe:  _root_.org.apache.spark.sql.execution.QueryExecution, exception:  scala.Exception): Unit = {

  }
}
```

3.3 tests give just per CTAS action.

```
A COMMAND JUST RAN
org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand
```

But 3.4 tests give this per CTAS action:

```
A COMMAND JUST RAN
org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand
A COMMAND JUST RAN
org.apache.spark.sql.execution.command.CreateDataSourceTableAsSelectCommand
```

This seems to support what I saw from printing out in LineageHarvester.

I unfortunately don't have the context to say whether this is ok or not :sweat_smile: . I would imagine Spline UI users would see this additional event, no?

The correct lineage events are getting generated but there's extra noise. I know InsertIntoHadoopFsRelation is associated with other writes occurring so your theory of Spark changing something under the hood is probably correct.

rycowhi: I think I found the [Spark PR](https://github.com/apache/spark/commit/4b40920e33176fc8b18380703e4dcf4d16824094) that introduced this behavior change in 3.4 - TLDR v1 data writes originally in CTAS turned into the two relation's we're seeing here under the two types of CTAS supported. They call out command nesting in various places.

integration-tests/src/test/scala/za/co/absa/spline/KafkaSinkSpec.scala
cerveada: Looks great, thanks

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
cerveada: So you are sure that both of those commands are triggered by `.write.mode(Append).saveAsTable(tableName)`?

This is not a question for testing only, but generally how to handle this in the application. Usually there is only one lineage for one write. The simplest solution would be to ignore one of them, but how to check if what we are ignoring is actually duplication?

cerveada: I read the Spark ticket, they actually separate the action of table creation and data insert, that is not a problem. We could generate events for both, but it should be clear that no data are inserted in the table creation lineage.

wajda:
@rycowhi Thanks a million for your effort.
Guys let me take a look at those event pairs more closely and I'll get back to you later. I'm a bit out of capacity at the moment, but will try to find some time during the next day or two.

integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
rycowhi: Would there be another suite/case that we'd want for something like this? I suppose that might depend on the actual events that Spline would be outputting here now.

devsaik:
Just checking on status of this PR. We are waiting on this feature. @wajda any updates.

wajda:
> Just checking on status of this PR. We are waiting on this feature. @wajda any updates.

Not yet, but it's high on my todo list for this week.

luke3481:
@wajda thanks for taking a look at @rycowhi's PR. Let me know if you'll be able to get to it this week - we're also waiting on this feature.

kaimalsreeja:
@wajda Looks like this feature is in high demand. Our team is also watching the PR merge, highly appreciate if you could review this as early as possible. Thank you in advance.

wajda:
Sorry folks, I've still been buried in work on another high-priority project, and have yet to obtain time allocation approval for this one. The Spline project was essentially put on hold by the company last year, that's the reason why it received almost no support for the past X months. I'm not giving up, but I can't really spend my free time on this any more, so trying to get some official allocation from the employer.

ramonje5:
> Sorry folks, I've still been buried in work on another high-priority project, and have yet to obtain time allocation approval for this one. The Spline project was essentially put on hold by the company last year, that's the reason why it received almost no support for the past X months. I'm not giving up, but I can't really spend my free time on this any more, so trying to get some official allocation from the employer.

Hello! Is there any news regarding this Spark 3.4 support? I've been using Spline since a long time and the fact that there is no support for Scala 2.13 is going to affect a lot of projects and maybe the need to find a new listener.

If there is any way I could contribute to this feature, or to speed things up, please let me know!

Thanks!

Edit: I've just seen [Upgrade org.scala-lang:scala-library from 2.12.17 to 2.13.14 #806](https://github.com/AbsaOSS/spline-spark-agent/pull/806)

wajda:
@ramonje5 thank you for the message. Unfortunately our company has yet to sort out its plans and book of work for Spline. Currently the team has zero capacity to work on it.

> If there is any way I could contribute to this feature, or to speed things up, please let me know!
> 
> Thanks!
> 
> Edit: I've just seen https://github.com/AbsaOSS/spline-spark-agent/pull/806

The PR #806 is automatically created by Snyk and won't work. Please ignore it. Upgrading to Scala 2.13 cannot be a simple change like that, it would require adding another vertical into the build matrix, so to speak. Basically the agent is built for every supported Sclala+Spark version combination.

If you want to help that would be awesome if you take this current PR, test it, address @rycowhi's points in the related issue #705. When testing succeeds and those points are addressed we can merge and release this PR.

rycowhi:
> @ramonje5 thank you for the message. Unfortunately our company has yet to sort out its plans and book of work for Spline. Currently the team has zero capacity to work on it.
> 
> > If there is any way I could contribute to this feature, or to speed things up, please let me know!
> > Thanks!
> > Edit: I've just seen #806
> 
> The PR #806 is automatically created by Snyk and won't work. Please ignore it. Upgrading to Scala 2.13 cannot be a simple change like that, it would require adding another vertical into the build matrix, so to speak. Basically the agent is built for every supported Sclala+Spark version combination.
> 
> If you want to help that would be awesome if you take this current PR, test it, address @rycowhi's points in the related issue #705. When testing succeeds and those points are addressed we can merge and release this PR.

At least for the versions maintained off of the existing cross-builds is there anything else to look into? I believe the only outstanding piece was seeing that there are now two (different commands) emitted off of a CTAS in 3.4 but have confirmed this is a new Spark behavior (by seeing the same, multiple events happen on a barebones query listener). Is the outstanding piece to see how that would look in the Spline UI, or?

integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
wajda:
```suggestion
          event.error.get.toString.toLowerCase should
            include(s"path ${tmpLocal.toURI.toString.stripSuffix("/")} already exists".toLowerCase)
```

wajda:
- Fixed a small issue (missing `toLowerCase` on the righthand side of the assert statement) in the test above.
- Added a cross-build config for Spark 3.4 + Scala 2.12 on the TeamCity.

When all builds pass I'll test it manually on my end and if no issue found I'll merge the PR.

PR #823, opened 2024-08-01
change to 9 parameters constructor

### LINKED ISSUE #821: Incompatible version between Spline and Spark BigQuery connector
I encounterd an exception as following when I tried to send Spark's lineage message to Kafka cluster.
```
24/07/29 22:56:26 ERROR org.apache.spark.util.Utils: throw uncaught fatal error in thread spark-listener-group-shared

java.lang.ExceptionInInitializerError

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$$anonfun$relationProviderProcessor$1.bigQueryConfig$1(BigQueryPlugin.scala:68)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$$anonfun$relationProviderProcessor$1.$anonfun$applyOrElse$7(BigQueryPlugin.scala:98)

    at scala.Option.orElse(Option.scala:447)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$$anonfun$relationProviderProcessor$1.applyOrElse(BigQueryPlugin.scala:98)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$$anonfun$relationProviderProcessor$1.applyOrElse(BigQueryPlugin.scala:62)

    at scala.PartialFunction$OrElse.apply(PartialFunction.scala:172)

    at za.co.absa.spline.harvester.plugin.composite.SaveIntoDataSourceCommandPlugin$$anonfun$writeNodeProcessor$1.applyOrElse(SaveIntoDataSourceCommandPlugin.scala:49)

    at za.co.absa.spline.harvester.plugin.composite.SaveIntoDataSourceCommandPlugin$$anonfun$writeNodeProcessor$1.applyOrElse(SaveIntoDataSourceCommandPlugin.scala:45)

    at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:176)

    at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:176)

    at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:176)

    at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:176)

    at scala.PartialFunction$Lifted.apply(PartialFunction.scala:228)

    at scala.PartialFunction$Lifted.apply(PartialFunction.scala:224)

    at za.co.absa.spline.harvester.builder.write.PluggableWriteCommandExtractor.asWriteCommand(PluggableWriteCommandExtractor.scala:45)

    at za.co.absa.spline.harvester.LineageHarvester.$anonfun$tryExtractWriteCommand$1(LineageHarvester.scala:145)

    at scala.util.Try$.apply(Try.scala:213)

    at za.co.absa.spline.harvester.LineageHarvester.tryExtractWriteCommand(LineageHarvester.scala:145)

    at za.co.absa.spline.harvester.LineageHarvester.harvest(LineageHarvester.scala:61)

    at za.co.absa.spline.agent.SplineAgent$$anon$1.$anonfun$handle$1(SplineAgent.scala:91)

    at za.co.absa.spline.agent.SplineAgent$$anon$1.withErrorHandling(SplineAgent.scala:100)

    at za.co.absa.spline.agent.SplineAgent$$anon$1.handle(SplineAgent.scala:72)

    at za.co.absa.spline.harvester.listener.QueryExecutionListenerDelegate.onSuccess(QueryExecutionListenerDelegate.scala:28)

    at za.co.absa.spline.harvester.listener.SplineQueryExecutionListener.$anonfun$onSuccess$1(SplineQueryExecutionListener.scala:41)

    at za.co.absa.spline.harvester.listener.SplineQueryExecutionListener.$anonfun$onSuccess$1$adapted(SplineQueryExecutionListener.scala:41)

    at scala.Option.foreach(Option.scala:407)

    at za.co.absa.spline.harvester.listener.SplineQueryExecutionListener.onSuccess(SplineQueryExecutionListener.scala:41)

    at org.apache.spark.sql.util.ExecutionListenerBus.doPostEvent(QueryExecutionListener.scala:155)

    at org.apache.spark.sql.util.ExecutionListenerBus.doPostEvent(QueryExecutionListener.scala:131)

    at org.apache.spark.util.ListenerBus.postToAll(ListenerBus.scala:117)

    at org.apache.spark.util.ListenerBus.postToAll$(ListenerBus.scala:101)

    at org.apache.spark.sql.util.ExecutionListenerBus.postToAll(QueryExecutionListener.scala:131)

    at org.apache.spark.sql.util.ExecutionListenerBus.onOtherEvent(QueryExecutionListener.scala:135)

    at org.apache.spark.scheduler.SparkListenerBus.doPostEvent(SparkListenerBus.scala:100)

    at org.apache.spark.scheduler.SparkListenerBus.doPostEvent$(SparkListenerBus.scala:28)

    at org.apache.spark.scheduler.AsyncEventQueue.doPostEvent(AsyncEventQueue.scala:37)

    at org.apache.spark.scheduler.AsyncEventQueue.doPostEvent(AsyncEventQueue.scala:37)

    at org.apache.spark.util.ListenerBus.postToAll(ListenerBus.scala:117)

    at org.apache.spark.util.ListenerBus.postToAll$(ListenerBus.scala:101)

    at org.apache.spark.scheduler.AsyncEventQueue.super$postToAll(AsyncEventQueue.scala:105)

    at org.apache.spark.scheduler.AsyncEventQueue.$anonfun$dispatch$1(AsyncEventQueue.scala:105)

    at scala.runtime.java8.JFunction0$mcJ$sp.apply(JFunction0$mcJ$sp.java:23)

    at scala.util.DynamicVariable.withValue(DynamicVariable.scala:62)

    at org.apache.spark.scheduler.AsyncEventQueue.org$apache$spark$scheduler$AsyncEventQueue$$dispatch(AsyncEventQueue.scala:100)

    at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.$anonfun$run$1(AsyncEventQueue.scala:96)

    at org.apache.spark.util.Utils$.tryOrStopSparkContext(Utils.scala:1381)

    at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.run(AsyncEventQueue.scala:96)

Caused by: java.lang.RuntimeException: Cannot find method `public static SparkBigQueryConfig from(... {7 args} ...)` in the class `class com.google.cloud.spark.bigquery.SparkBigQueryConfig`

    at scala.sys.package$.error(package.scala:30)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$SparkBigQueryConfig$.$anonfun$methodFrom$2(BigQueryPlugin.scala:150)

    at scala.Option.getOrElse(Option.scala:189)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$SparkBigQueryConfig$.<init>(BigQueryPlugin.scala:150)

    at za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin$SparkBigQueryConfig$.<clinit>(BigQueryPlugin.scala)

    ... 47 more
```

Then, I went to check the commits history records of [BigQueryPlugin.scala](https://github.com/AbsaOSS/spline-spark-agent/blob/develop/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala#L150). I found that there has always been a check for whether the number of parameter types equals 7, right after the initial commit.

Hence, base on the prompt, I go to SparkBigQueryConfig.java and check the commits history records as following:
[v0.25.0](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/f1f7b2ddd96ff3596034f68d7d2c0df6ee968f21/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java#L180)
[v0.25.1](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/c6487a8e721c83e630b82fec53a4f373df23aa4f/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java#L187)

Base on the above commits history, it indeed provides a static SparkBigQueryConfig with 7 args  before version 0.25.0, but it was changed after 0.25.0.
Therefore, you will encounter this error if you use the version after this version.

XavierZYXue:
fixes #821 to switch the 9 parameters constructor, could you please review it?

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
@@ -16,7 +16,12 @@
 
 package za.co.absa.spline.harvester.plugin.embedded
 
+import java.lang.reflect.Method
+import java.lang.reflect.Modifier.isStatic
+import java.util.Optional
+
 import io.github.classgraph.ClassGraph
+import javax.annotation.Priority
 import org.apache.hadoop.conf.Configuration
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.execution.datasources.{LogicalRelation, SaveIntoDataSourceCommand}
@@ -31,10 +36,6 @@ import za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin.SparkBigQueryC
 import za.co.absa.spline.harvester.plugin.embedded.BigQueryPlugin._
 import za.co.absa.spline.harvester.plugin.{BaseRelationProcessing, Plugin, RelationProviderProcessing}
 
-import java.lang.reflect.Method
-import java.lang.reflect.Modifier.isStatic
-import java.util.Optional
-import javax.annotation.Priority
 import scala.collection.JavaConverters._
 import scala.language.reflectiveCalls
 
@@ -142,14 +143,39 @@ object BigQueryPlugin {
       def getParentProjectId: String
     }
     private val clazz = findPossiblyShadedClass("com.google.cloud", "com.google.cloud.spark.bigquery.SparkBigQueryConfig")
-    private val methodFrom: Method = clazz
+    val from: (java.util.Map[_, _], ImmutableMap.ImmutableMap, Configuration, ImmutableMap.ImmutableMap, Integer, SQLConf, String, Optional[StructType], java.lang.Boolean) => SparkBigQueryConfig = {
+      case (a, b, c, d, e, f, g, h, i) =>
+        if (methodFrom7.isDefined)
+          methodFrom7.get.invoke(clazz, a, b, c, e, f, g, h).asInstanceOf[SparkBigQueryConfig]
+        else if (methodFrom8.isDefined)
+          methodFrom8.get.invoke(clazz, a, b, c, e, f, g, h, i).asInstanceOf[SparkBigQueryConfig]
+        else
+          methodFrom9.get.invoke(clazz, a, b, c, d, e, f, g, h, i).asInstanceOf[SparkBigQueryConfig]
+    }
+    private val methodFrom7: Option[Method] = clazz
+      .getMethods
+      .find(
+        m => m.getName == "from"
+          && isStatic(m.getModifiers)
+          && m.getParameterTypes.length == 7
+          && m.getReturnType.getSimpleName == "SparkBigQueryConfig"
+      )
+    private val methodFrom8: Option[Method] = clazz
+      .getMethods
+      .find(
+        m => m.getName == "from"
+          && isStatic(m.getModifiers)
+          && m.getParameterTypes.length == 8
+          && m.getReturnType.getSimpleName == "SparkBigQueryConfig"
+      )
+    private val methodFrom9: Option[Method] = clazz
       .getMethods
       .find(
         m => m.getName == "from"
           && isStatic(m.getModifiers)
           && m.getParameterTypes.length == 9
           && m.getReturnType.getSimpleName == "SparkBigQueryConfig"
-      ).getOrElse(sys.error(s"Cannot find method `public static SparkBigQueryConfig from(... {9 args} ...)` in the class `$clazz`"))
+      )
 
     object ImmutableMap {
       type ImmutableMap = AnyRef
@@ -166,10 +192,10 @@ object BigQueryPlugin {
           .asInstanceOf[ImmutableMap]
     }
 
-    val from: (java.util.Map[_, _], ImmutableMap.ImmutableMap, Configuration, ImmutableMap.ImmutableMap, Integer, SQLConf, String, Optional[StructType], java.lang.Boolean) => SparkBigQueryConfig =
-      methodFrom
-        .invoke(clazz, _, _, _, _, _, _, _, _, _)
-        .asInstanceOf[SparkBigQueryConfig]
+    private val methodFrom: Method = methodFrom7
+      .orElse(methodFrom8)
+      .orElse(methodFrom9)
+      .getOrElse(sys.error(s"Cannot find method `public static SparkBigQueryConfig from(... {7|8|9 args} ...)` in the class `$clazz`"))
   }
 
   private object `_: DirectBigQueryRelation` extends SafeTypeMatchingExtractor[AnyRef]("com.google.cloud.spark.bigquery.direct.DirectBigQueryRelation")
--- a/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
@@ -167,15 +167,10 @@ object BigQueryPlugin {
           && m.getParameterTypes.length == 9
           && m.getReturnType.getSimpleName == "SparkBigQueryConfig"
       )
-    val from: (java.util.Map[_, _], ImmutableMap.ImmutableMap, Configuration, ImmutableMap.ImmutableMap, Integer, SQLConf, String, Optional[StructType], java.lang.Boolean) => SparkBigQueryConfig = {
-      case (a, b, c, d, e, f, g, h, i) =>
-        if (methodFrom7.isDefined)
-          methodFrom7.get.invoke(clazz, a, b, c, e, f, g, h).asInstanceOf[SparkBigQueryConfig]
-        else if (methodFrom8.isDefined)
-          methodFrom8.get.invoke(clazz, a, b, c, e, f, g, h, i).asInstanceOf[SparkBigQueryConfig]
-        else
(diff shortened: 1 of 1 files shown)


cerveada:
Hello, thanks for your contribution.

Please correct me if I am wrong, but this fixes the issues for the new version of BigQuery, but also it will break it for the older versions.

Another thing I notice is that in the linked Big Query code the method has 8 params, but here you use 9. How come?

XavierZYXue:
@cerveada hi, first of appreciated your response, I revised this logic because it was changed to 9 parameters after this [PR](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/commit/360e199aba9b72593f52f52adc781a714346cafc#diff-793e3bca4e47599c6842972e09059c9a357006c7c2dc9e6202647576073c619a), and affetced all versions after 0.26.0
that is to say, it will be applicable to all versions after 0.26.0

cerveada:
That is ok, but we should keep it compatible with the old version as well. We already do this in other plugins. Usually some combination of retries and reflection is sufficient to support both.

You can see some examples of it here:
core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/DataSourceV2Plugin.scala

XavierZYXue:
@cerveada I've added different constructors separately are compatible with all Spark BigQuery connector
please review again?

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
```diff
@@ -157,12 +180,22 @@ object BigQueryPlugin {
...
-        .asInstanceOf[SparkBigQueryConfig]
+    val from: (java.util.Map[_, _], ImmutableMap.ImmutableMap, Configuration, ImmutableMap.ImmutableMap, Integer, SQLConf, String, Optional[StructType], java.lang.Boolean) => SparkBigQueryConfig = {
+      case (a, b, c, d, e, f, g, h, i) =>
+        if (methodFrom7.isDefined)
+          methodFrom7.get.invoke(clazz, a, b, c, e, f, g, h).asInstanceOf[SparkBigQueryConfig]
+        else if (methodFrom8.isDefined)
+          methodFrom8.get.invoke(clazz, a, b, c, e, f, g, h, i).asInstanceOf[SparkBigQueryConfig]
+        else
+          methodFrom9.get.invoke(clazz, a, b, c, d, e, f, g, h, i).asInstanceOf[SparkBigQueryConfig]
+    }
```
wajda: This piece of code seems over-complicated to me. I've got a couple of points:
1. Why are you using a partial function? Won't just a normal function do?
2. I guess we need to clearly define all three method signatures with proper parameter names for clarity.
3. To me, there's no reason to create `Option[Method]` and then de-reference it via `isDefined-get`. That is improper usage of the `Option` monad. I would recommend to choose between the following two alternative approaches:

a) Replace `methodFromN: Option[Method]` with `invokeMethodFromN: Option[(... 9-arg signature ...) => ReturnValue`, chain them with `orElse` and call the chain here like `invokeMethodFrom(1, 2, 3, ,,,, 9)`, or

b) (probably even better approach) Do no use `Option` at all. Lookup any `from` method, and then use a pattern matching on the signature to decide how the method should be be called. Consider this pseudo-code:
```scala
val methodFrom: Method = clazz
      .getMethods
      .find(
        m => m.getName == "from"
          && isStatic(m.getModifiers)
          && m.getReturnType.getSimpleName == "SparkBigQueryConfig"
      )
      .getOrElse(sys.error(s"Cannot find any "from(...7|8|9...)" method....")
      
...

methodFrom.getParameterTypes.length match {
  case 7:  invoke(clazz, ...{7}...)
  case 8:  invoke(clazz, ...{8}...)
  case 9:  invoke(clazz, ...{9}...)
  case x => sys.error(s"Unsupported method signature ......")
}
```

wajda (changes requested):
The build for the Spark 3.4 is failing due to timeout.
Also please see my comment below.

core/src/main/scala/za/co/absa/spline/harvester/plugin/embedded/BigQueryPlugin.scala
XavierZYXue: yes, good advise for me, I've changed. please check again.

XavierZYXue: I only need to care whether one of the three methods 7|8|9 exists, and then invoke the corresponding method according to the number of parameters.
Instead of having to define the three methods separately.

XavierZYXue:
any comment for this after I changed? and why the pipeline always timeout? do you know the reason

wajda (approved):
LGTM.
Regarding timeout - I think there is some non-deterministic behaviour in tests. It also happens to other branches, so no relation to this PR.

cerveada (approved):
LGTM

rkrumins:
Amazing, thank you very much @cerveada and @cerveada

PR #924, opened 2026-07-21
issue #922 Fix lineage crash on Column.withField/dropFields

### LINKED ISSUE #922: withField/dropFields crashes harvesting with "StructFieldsOperation.dataType should not be called"
Any job that uses Column.withField(...) or dropFields(...) fails during lineage harvest. 
The job itself finishes fine (SplineAgent swallows the error), but lineage for that execution is dropped entirely.

Repro:

```
import org.apache.spark.sql.functions._

val df = spark.range(1)
  .select(struct(lit(1).as("a"), lit(2).as("b")).as("s"))
  .withColumn("s", col("s").withField("a", lit(99)))

df.write.mode("overwrite").saveAsTable("tmp.repro_table")

```

Error:

```
ERROR agent.SplineAgent: Unexpected error occurred during lineage processing for application: ...
java.lang.IllegalStateException: StructFieldsOperation.dataType should not be called.
	at org.apache.spark.sql.catalyst.expressions.complexTypeCreator$StructFieldsOperation.dataType(...)
	at za.co.absa.spline.harvester.converter.ExpressionConverter.convertDataType(ExpressionConverter.scala)
	at za.co.absa.spline.harvester.converter.ExprToRefConverter.convert(ExprToRefConverter.scala:42)
	...
```

WithField/DropField are children of UpdateFields, and Spark deliberately leaves their dataType unimplemented — only the parent folds them to compute the result type. ExpressionConverter calls .dataType on every node uniformly, so it hits this guard.

Same shape as the WindowSpecDefinition/WindowFrame case already handled here:

```
case _: sparkExprssions.WindowSpecDefinition
     | _: sparkExprssions.WindowFrame =>
  FunctionalExpression(dataType = None, ...)

```

Adding StructFieldsOperation (or WithField/DropField directly) to that same match seems like it'd fix it. Can send a PR if that's the right approach.

Env: spline-spark-agent 2.2.3 (spark-3.4-spline-agent-bundle_2.12), Spark 3.4.2, Scala 2.12

LeeChaeRok:
Fixes #922

Column.withField(...) / dropFields(...) create WithField/DropField expressions, which are children of UpdateFields (Spark 3.1+). 
Spark leaves dataType unimplemented on these two on purpose, only UpdateFields folds them to get the real result type. ExpressionConverter.convert calls .dataType on every expression node without exception, so it hits this guard and throws whenever a job uses withField/dropFields.

Since SplineAgent swallows the exception, the job itself finishes fine but lineage for that execution is dropped entirely, which is worse than an error because it just fails silently.

I didn't match on the type directly (case _: StructFieldsOperation) like the existing WindowSpecDefinition/WindowFrame case does, because StructFieldsOperation and its subtypes don't exist before Spark 3.1, and core compiles as one source tree against spark-2.2 ~ spark-3.5 profiles. So matched by class name instead (e.getClass.getSimpleName), this way it still compiles on the older profiles too.

Added two unit tests in ExpressionConverterSpec for WithField/DropField, checking dataType stays None and dtConverter is never called.

Tested locally on spark-3.4 profile, all existing tests plus the 2 new ones pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Spark field-manipulation expressions, including `WithField` and `DropField`.
  * These expressions are now recorded as untyped while preserving applicable child-expression references, improving compatibility across Spark versions.

### DIFF
--- a/core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
+++ b/core/src/main/scala/za/co/absa/spline/harvester/converter/ExpressionConverter.scala
@@ -101,6 +101,16 @@ class ExpressionConverter(
         params = getExpressionParameters(sparkExpr)
       )
 
+    case e if UntypedExpressionClassNames(e.getClass.getSimpleName) =>
+      FunctionalExpression(
+        id = idGen.nextId(),
+        dataType = None,
+        childRefs = convertChildren(e),
+        extra = createExtra(e, ExprV1.Types.UntypedExpression),
+        name = e.prettyName,
+        params = getExpressionParameters(e)
+      )
+
     case e: sparkExprssions.Expression =>
       FunctionalExpression(
         id = idGen.nextId(),
@@ -148,6 +158,9 @@ object ExpressionConverter {
     }
 
   }
+
+  // matched by name, not type, to keep compiling against Spark < 3.1 where these classes don't exist
+  private val UntypedExpressionClassNames: Set[String] = Set("WithField", "DropField")
 
   private val IgnoredSparkExprPropNames: Set[String] = Set("children", "dataType", "nullable")
   private val IgnoredSparkExprPropTypes: Set[Class[_]] = Set(classOf[ExprId])
--- a/core/src/test/scala/za/co/absa/spline/harvester/converter/ExpressionConverterSpec.scala
+++ b/core/src/test/scala/za/co/absa/spline/harvester/converter/ExpressionConverterSpec.scala
@@ -28,7 +28,7 @@ import org.scalatest.{Inside, OneInstancePerTest}
 import org.scalatestplus.mockito.MockitoSugar
 import za.co.absa.spline.harvester.SequentialIdGenerator
 import za.co.absa.spline.model.dt
-import za.co.absa.spline.producer.model.FunctionalExpression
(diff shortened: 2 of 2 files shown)


wajda:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/wajda/pr-checks/5e7b3884-09c8-4296-a0f7-5b2ca27f36ae?refs=791bfa89f897547cf77b7bbbc4c412560bc08041&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

wajda (approved):
LGTM

